Issue #19, opened 2022-12-12
ProviderSettingsExport.json is broken JSON


Sankgreall:
# 🐛 Summary #
ProviderSettingsExport.json is broken JSON, resulting in the following reporting error:

```
ConvertFrom-Json : Invalid JSON primitive: .
At C:\Users\Sankgreall\Documents\AzureDevOps\ScubaGear\PowerShell\ScubaGear\Modules\CreateReport\CreateReport.psm1:39
char:44
+ $SettingsExport =  Get-Content $FileName | ConvertFrom-Json
+                                            ~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [ConvertFrom-Json], ArgumentException
    + FullyQualifiedErrorId : System.ArgumentException,Microsoft.PowerShell.Commands.ConvertFromJsonCommand
```

Upon investigating ProviderSettingsExport.json and sticking it in a JSON validator, it flagged the following section:

```
            "conditional_access_policies": ,
    "authorization_policies": {
    "AllowEmailVerifiedUsersToJoinOrganization":  false,
    "AllowInvitesFrom":  "everyone",
    "AllowUserConsentForRiskyApps":  null,
    "AllowedToSignUpEmailBasedSubscriptions":  true,
    "AllowedToUseSspr":  true,
    "BlockMsolPowerShell":  false,
    "DefaultUserRoleOverrides":  null,
    "DefaultUserRolePermissions":  {
                                       "AllowedToCreateApps":  true,
                                       "AllowedToCreateSecurityGroups":  true,
                                       "AllowedToReadBitlockerKeysForOwnedDevice":  true,
                                       "AllowedToReadOtherUsers":  true
                                   },
    "DeletedDateTime":  null,
    "Description":  "Used to manage authorization related settings across the company.",
    "DisplayName":  "Authorization Policy",
    "EnabledPreviewFeatures":  [

                               ],
    "GuestUserRoleId":  "10dae51f-b6af-4016-8d66-8c2a99b929b3",
    "Id":  "authorizationPolicy",
    "PermissionGrantPolicyIdsAssignedToDefaultUserRole":  [
                                                              "ManagePermissionGrantsForSelf.microsoft-user-default-legacy"
                                                          ],
    "AdditionalProperties":  {

                             }
},
```

I'm assuming this is either supposed to be `"conditional_access_policies": "[some value]"` or `"conditional_access_policies": {... nested dict}`

## To reproduce ##

[...]

PR #21, opened 2022-12-13
Empty cond access validation

Sankgreall:
- Added a conditional check to verify if the result of the conditional access policy check was null

- If null, set variable to an empty array so that it doesn't create invalid JSON when outputted to file

Refer to issue #19 

## 🧪 Testing ##

- Run the ./SCuBA.ps1 against an estate with no conditional access policies
- Open ProviderSettingsExport.json once complete
- Check the key `conditional_access_policies`
- Correct value should be `[]`
- Incorrect value will likely throw the error below to console
```
ConvertFrom-Json : Invalid JSON primitive: .
At C:\Users\Sankgreall\Documents\AzureDevOps\ScubaGear\PowerShell\ScubaGear\Modules\CreateReport\CreateReport.psm1:39
char:44
+ $SettingsExport =  Get-Content $FileName | ConvertFrom-Json
+                                            ~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [ConvertFrom-Json], ArgumentException
    + FullyQualifiedErrorId : System.ArgumentException,Microsoft.PowerShell.Commands.ConvertFromJsonCommand
```

      in code comments.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Add a tag or create a release.

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
@@ -19,6 +19,14 @@ function Export-AADProvider {
         # - 2.17 first part
         $AllPolicies = Get-MgIdentityConditionalAccessPolicy -ErrorAction Stop | ConvertTo-Json -Depth 10
 
+        # Set to empty array if returned value is null
+        if ($null -eq $AllPolicies) {
+            # Empty arrays are mangled into $null,
+            # However, you can preserve an array on
+            # return by prepending it with the array construction operator ","
+            $AllPolicies = ,@() | ConvertTo-Json
+        }
+
         # Get a list of the tenant's provisioned service plans - used to see if the tenant has AAD premium p2 license required for some checks
         # The Rego looks at the service_plans in the JSON
         $ServicePlans = (Get-MgSubscribedSku).ServicePlans | Where-Object -Property ProvisioningStatus -eq -Value "Success" -ErrorAction Stop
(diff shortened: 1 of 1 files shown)


PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -19,6 +19,12 @@ function Export-AADProvider {
         # - 2.17 first part
         $AllPolicies = Get-MgIdentityConditionalAccessPolicy -ErrorAction Stop | ConvertTo-Json -Depth 10
 
+        # Set to empty array if returned value is null
+        if ($null -eq $AllPolicies) {
+            # Assign an empty array
+            $AllPolicies = ConvertTo-Json @()
+        }
+
```
gdasher, line 27: Thanks for the change.  We're actually going to make this change across the board (not just when the policies are null); see for example the new version of the code that just landed today in ExportEXOProvider (we just couldn't squeeze in AAD unfortunately).

So if its ok by you, we would prefer to delay this PR until we make that change.  We're also planning to move our internal development into this public repo soon rather than dropping bundles of code, so it shouldn't be too long until we land a version of this at HEAD.

Sankgreall: Yeah no worries. Just thought it would be rude not to submit a PR when the issue was so simple. Feel free to abandon if you prefer.

nanda-katikaneni:
Thanks for the PR and suggested changes - the changes were merged through another PR as part of our larger AAD provider error handling and to fix a few earlier issues: #83 #68 #19. 
Closing this PR.

Issue #68, opened 2022-12-22
Add Error-handling for AAD provider


schrolla:
Decomposed item from #207 (look into that issue for more details)

PR #100, opened 2023-01-04
Add AAD Error Handling

### LINKED ISSUE #83: ConvertTo-Json for "AAD"
modify all ConvertTo-Json in AAD provider script to $Output = ConvertTo-Json @(Get-ExampleCommand) for consistency and test to make sure the changes dont break the current code
related to #197

adhilto:
## 🗣 Description ##
- Updated all of the ConvertTo-Json calls to use the current syntax
- Wrapped all API calls inside the `TryCommand` function
- Updated Rego so that it always expects provider output items to be lists

Closes #83, #68, and #19.

## 🧪 Testing ##
- Ran code against Test Tenant 1 (G5) on both main and this branch, closely comparing output and ensuring that all differences are expected.
- Ran code against Test Tenant 2 (G3) on both main and this branch, closely comparing output and ensuring that all differences are expected.
- Ran code in Test Tenant 4 (E5) as an unprivileged user to generate errors. Ensured that all errors generated were handled correctly.
- Ran code in Test Tenant 4 (E5) as an admin after deleting all conditional access policies. Was able to reproduce the error described in #19 in main; bug is not present on new branch.

      in code comments.
- [x] ~~I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.~~

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Add a tag or create a release.

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
@@ -23,7 +23,7 @@ function Export-AADProvider {
 
     # Get a list of the tenant's provisioned service plans - used to see if the tenant has AAD premium p2 license required for some checks
     # The Rego looks at the service_plans in the JSON
-    $ServicePlans = $Tracker.TryCommand("Get-MgSubscribedSku", @{"fakeparm"="hi"}).ServicePlans | Where-Object -Property ProvisioningStatus -eq -Value "Success"
+    $ServicePlans = $Tracker.TryCommand("Get-MgSubscribedSku").ServicePlans | Where-Object -Property ProvisioningStatus -eq -Value "Success"
 
     if ($ServicePlans) {
         # The RequiredServicePlan variable is used so that PIM Cmdlets are only executed if the tenant has the premium license
--- a/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
@@ -63,6 +63,9 @@ function Export-AADProvider {
     # 2.7 Policy Bullet 2]
     $AdminConsentReqPolicies = ConvertTo-Json @($Tracker.TryCommand("Get-MgPolicyAdminConsentRequestPolicy"))
 
+    $SuccessfulCommands = ConvertTo-Json @($Tracker.GetSuccessfulCommands())
+    $UnSuccessfulCommands = ConvertTo-Json @($Tracker.GetUnSuccessfulCommands())
+
     # Note the spacing and the last comma in the json is important
     $json = @"
     "conditional_access_policies": $AllPolicies,
@@ -72,6 +75,8 @@ function Export-AADProvider {
     "privileged_roles": $PrivilegedRoles,
     "service_plans": $ServicePlans,
     "directory_settings": $DirectorySettings,
+    "aad_successful_commands": $SuccessfulCommands,
+    "aad_unsuccessful_commands": $UnSuccessfulCommands,
 "@
 
     # We need to remove the backslash characters from the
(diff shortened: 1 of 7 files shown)


tkol2022:
@nanda-katikaneni Please regression test this using the new semi-automated test script against Test Tenant 5 (E3) for AAD only.

- [ ] Generate golden copy of provider and test results using main branch code 
- [ ] Run semi-automated test script in the "add-error-handling" branch using Rego test option to make sure outputs are the same
- [ ] Run ScubaGear using the "add-error-handling" code and make sure the provider functions as expected

rgbrow1949:
I compared the report results for my CISA account, my test account, and an account with no permissions on the main branch and on this branch. 

This account correctly displays Errors on the report in all cases, all unit tests pass, and the JSON outputs are correct as well without redundancies. 

I approve.

rgbrow1949 (approved):
I compared the report results for my CISA account, my test account, and an account with no permissions on the main branch and on this branch.

This account correctly displays Errors on the report in all cases, all unit tests pass, and the JSON outputs are correct as well without redundancies.

I approve.

tkol2022:
> @tkol2022 I discovered a bug. You can see if by running `Disconnect-MgGraph` then the tool with `-Login $false` image
> 
> The issue is with how I handled $PrivilegedUsers. image Everything said in the comment is correct...but on error, `$TryCommand` returns an empty list, not a dict like the function normally does, resulting in invalid json.
> 
> I don't want to interrupt your testing. Should I fix this now or wait until you finish?

You can fix it now. Should not affect my testing. Thanks!

tkol2022:
###  Not sure if this is a bug?
So I know you mentioned at today's meeting that there are some cases where a policy will be missing from the report if its respective provider Cmdlet fails but I can't remember the full context of what you said, so you may know about this already.

When the Get-MgDirectorySetting errors, the Baseline 2.7: Policy 3 "Group owners SHALL NOT be allowed to consent to third-party applications" gets excluded from the report.

image
image

tkol2022:
### Might need to add logic to handle the scenario where Get-MgSubscribedSku fails
If Get-MgSubscribedSku fails, we should not call Get-PrivilegedUser and Get-PrivilegedRole because those functions depend on the list of SKUs to determine whether or not they should execute PIM-specific logic. Both of those functions have the line of code below:

  `  if ($TenantHasPremiumLicense) {`

You might need to add an if block to ensure that those functions are only called if Get-MgSubscribedSku succeeds. See my example below.

```
    $ServicePlans = $Tracker.TryCommand("Get-MgSubscribedSku").ServicePlans | Where-Object -Property ProvisioningStatus -eq -Value "Success" -ErrorAction Stop
    if($ServicePlans){
       ### Put all code that calls Get-PrivilegedUser and Get-PrivilegedUser HERE
    }

```

tkol2022 (changes requested):
See comments. I will also test the error handling of Get-PrivilegedUser once you commit the fix you mentioned.

adhilto:
> ### Not sure if this is a bug?
> So I know you mentioned at today's meeting that there are some cases where a policy will be missing from the report if its respective provider Cmdlet fails but I can't remember the full context of what you said, so you may know about this already.
> 
> When the Get-MgDirectorySetting errors, the Baseline 2.7: Policy 3 "Group owners SHALL NOT be allowed to consent to third-party applications" gets excluded from the report.
> 
> image image

Good catch. I thought this was worth fixing, so I just pushed a patch for it. We should fix all cases that we know cause something to be missing from the report. What I was saying in the meeting you referred to was that if there ever were something missing from the report (e.g., due to a Rego bug), we don't currently have code in place to flag that for the user, so unless catches it like you did, it will go unnoticed.

adhilto:
> 

Added the logic you suggested.

@schrolla All requested changes have been made.

nanda-katikaneni (approved):
Tested main branch and the aad-error-handling branches on E3 tenant - results look fine. 
Note: that the aad-error-handling branch has an issue with automation code (the issue is since fixed in main through a separate merge) - so did tests with modified automation-script. The changes made in this branch looks good and does not conflict with automation changes - approving.

PR #333, opened 2023-05-12
Teams email integration patch

adhilto:
## 🗣 Description ##
Teams email integration is only available in commercial environments. However, at least one user has reported the tool falsely reporting email integration as enabled in a GCC environment and existing code did not probably handle GCC-high tenants.

Changes made:
- Expose the $M365Environment setting to the Rego
- Adjust the Teams Rego to automatically pass the team email integration check if $M365Environment is not set to "commercial".
- Adjust the Teams unit tests for this change

## 💭 Motivation and context ##

Closes #259.

## 🧪 Testing ##
- Tested on E5 environment with email integration enabled with $M365Environment set to "commercial"
- Tested on E5 environment with email integration disabled with $M365Environment set to "commercial"
- Tested on E5 environment with email integration enabled but $M365Environment set to "gcc"
- Tested on G5 environment with $M365Environment set to "gcc"
- Tested on G5 environment with $M365Environment set to "commercial"
- Tested on GCC high tenant

## ✅ Pre-approval checklist ##
- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] *All* future TODOs are captured in issues, which are referenced
      in code comments.
- [x] All relevant type-of-change labels have been added.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All relevant repo and/or project documentation has been updated
      to reflect the changes in this PR.
- [x] Tests have been added and/or modified to cover the changes in this PR.
- [x] All new and existing tests pass.

## ✅ Pre-merge checklist ##

- [ ] Revert dependencies to default branches.
- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Create a release.

PR #397, opened 2023-06-21
Teams baseline update based on the pilot feedback and internal testing

### LINKED ISSUE #241: Adjudicate Teams comments on the comment matrix
- [ ] Send out link to Comments Matrix to the Baseline POCs
- [ ] Andrew to schedule meetings with Baseline POCs
- [ ] Internally review (Andrew & Baseline POCs) Teams comments from the comments matrix 
- [ ] Review comments with CISA
- [ ] Open GitHub Issues as needed
- [ ] Prioritize GitHub Issues into release lifecycle

### LINKED ISSUE #260: Revise Teams policy 2.4 based on the feedback and internal testing
# 💡 Summary #

Modify M365 Teams baseline policy 2.4 as per the instructions below.

## Motivation and context ##

Based on the pilot and internal feedback, the existing policy has ambiguity and/or buttonology that can potentially change.

## Implementation notes ##

- [x] Ambiguity in baseline text. Recommend adding "in the Global (Org-wide default) meeting policy or in custom meeting policies if any exist."
- [x] Clarify external user access and "allow"restriction"; define "guest" 
- [x] The policy bullet uses "buttonology." Recommend changing the policy bullets to the following:
"Only the meeting organizer SHOULD be able to record live events in the global (org-wide default) live events policy.
Alternate live events policies MAY be created that allow agency-approved users the ability to record."

## Acceptance criteria ##

Updated policy has the changes noted in Implementation notes.

### LINKED ISSUE #267: Teams 2.8 Baseline Markdown Update
# 💡 Summary #

Teams 2.8 policy states "Agencies SHOULD allow all apps published by Microsoft, but MAY block specific Microsoft apps as needed", "Agencies SHOULD NOT allow installation of all third-party apps or custom apps, but MAY allow specific apps as needed" and "Agencies SHALL establish policy dictating the app review and approval process to be used by the agency".
Alden has noted that there is ambiguity on the scope of the policies.

## Motivation and context ##

Policy language and implementation update to clearly state the expected policies for teams to be scoped to the Global (org-wide default) meeting policies and all custom policies.

## Implementation notes ##

- Andrew to update the policy in the baseline to clearly state for these policies to be coped to the Global (org-wide default) meeting policies and all custom policies.
- Check and update the rego as needed to ensure the code is checking the Global (org-wide default) meeting policies and all custom policies. 
- Policies containing “MAY” will be removed due to current capabilities of Scubagear being able to check the policy being enabled in only the global and not custom policies.
- Add note to policy stating "Custom app policies MAY be created to allow more flexibility for specific users”

## Acceptance criteria ##

Changes suggested in implementation notes above:

- [x]  baseline language is updated to to scope policy to be scoped to the Global (org-wide default) meeting policies and all custom policies.
- [x] rego is checking the policy against the Global (org-wide default) meeting policies and all custom policies.
- [x] "MAY" policies are removed
- [x] "Custom app policies MAY be created to allow more flexibility for specific users” note added

### LINKED ISSUE #268: Teams 2.10 Baseline Markdown Update
# 💡 Summary #

Teams 2.10 policy states "Only the Meeting Organizer SHOULD Be Able to Record Live Events"
Alden has noted that there is ambiguity on the scope of the policies.

## Motivation and context ##

olicy language and implementation update to clearly state the expected policies for teams to be scoped to the Global (org-wide default) meeting policies and all custom policies.

## Implementation notes ##

- Andrew to update the policy in the baseline to clearly state for these policies to be coped to the Global (org-wide default) meeting policies and all custom policies.
- Check and update the rego as needed to ensure the code is checking the Global (org-wide default) meeting policies and all custom policies. 
- Policies containing “MAY” will be removed due to current capabilities of Scubagear being able to check the policy being enabled in only the global and not custom policies.
- Add note to policy stating "Custom app policies MAY be created to allow more flexibility for specific users”

## Acceptance criteria ##

Changes suggested in implementation notes above:

- [x]  baseline language is updated to to scope policy to be scoped to the Global (org-wide default) meeting policies and all custom policies.
- [x] rego is checking the policy against the Global (org-wide default) meeting policies and all custom policies.
- [x] "MAY" policies are removed
- [x] "Custom app policies MAY be created to allow more flexibility for specific users” note added

### LINKED ISSUE #279: Teams Policy 2.6 - update to clarify the support for Skype
# 💡 Summary #

Teams Policy 2.6 states "Microsoft officially retired Skype for Business Online on July 31, 2021, and it is no longer supported". As OPM has highlighted Skype for business support is been extended to 2025; check to see if the policy / rego is scoped to cover Skype online and Skype for business and make any necessary updates. 

## Motivation and context ##

Policy statement update based on new end of support timeline for Skype for Business Online. Based on the policy update, it may require changes to Rego policy, in which case a separate issue will be added. 

## Implementation notes ##

Update the policy statement to provide new end of life support for Skype for Business Online and if this require a change in pass/fail criteria.

## Acceptance criteria ##

Changes suggested in implementation notes above.

### LINKED ISSUE #363: Teams Policy 3 (MS.TEAMS.3.1v1) - 'Implementation' steps need to be revised
# 💡 Summary #

Modify M365 Teams baseline policy 3 as per the instructions below.

Motivation and context
Based on the pilot and internal feedback, the existing  implementation steps are ambiguous and needs a revision.

Implementation notes
The Implementation section step #4, suggest adding a recommendation for what it SHOULD be set to. Possible language for this: "... ensure Automatically admit people is NOT set to Everyone. Automatically admit people" SHOULD be set to "People in my organization", though other options MAY be used if needed.
 
Acceptance criteria
Updated policy has the changes noted in Implementation notes.

### LINKED ISSUE #383: Teams third party apps restriction not producing accurate results in certain scenarios
# 🐛 Summary #

The ScubaGear MS Teams policy 2.8 or MS.TEAMS.8.2v1 "Agencies SHOULD NOT allow installation of all third-party apps or custom apps, but MAY allow specific apps as needed" **produces a Fail (non-compliant) in certain scenarios when it should produce a Pass**.

_Note: This issue describes the problem for the "third-party apps" setting but I believe the same problem could occur for the "custom apps" setting as well. You should test that and incorporate the fix if needed._

_Also, the terminology regarding the different settings screens can be confusing. Apparently there are two separate screens: the **Org-wide app settings** screen and the **Global (Org-wide default) policy settings** screen. Read the expected behavior for details on the two screens._

## To reproduce ##

1. Go to Teams Admin portal > Teams apps > Permission policies
2. Open the Global (Org-wide default) policy
3. Change the Third-party apps option to Allow all apps (you might need to press Save at this step to reproduce the behavior - not sure)
4. Click the Org-wide app settings button at the top (I noticed in commercial tenants you need to navigate to Teams apps > Manage apps to see the Org-wide app settings button).
5. Under Third-party apps set the configuration to Off
6. Click save on the Org-wide app settings screen
7. Click save again on the Global (Org-wide default) policy page
8. Open Teams as a test user and you will see that you cannot install third-party apps
9. Run ScubaGear and you will see that the code produces a fail (Warning) in the report

## Expected behavior ##

The **Org-wide app settings** for Teams take priority over the **Global (Org-wide default) policy settings**. Therefore since the Org-wide app setting is set to Off (i.e. block third party apps), then it does not matter what the data value is for the Global (Org-wide default) policy setting. Teams will enforce the blocking of third party apps. I verified via hands-on testing.

Since this is the case, I would expect ScubaGear to produce a Pass (requirement met = true) in this scenario.

## Potential Problem in the Code ##

I think what might be happening is that the code is not looking at the Org-wide app setting. It is only looking at the Global (Org-wide default) policy setting.

## Screenshots ##

See below - The installation of third-party apps is disabled in the Org-wide app setting and therefore overrides the Global (Org-wide default) policy setting
image

See below - Teams is enforcing the "no third-party apps" restriction because all of the third party apps that were previously installed disappeared and I was unable to install any more.
image

See below - This was the value set in the Global (Org-wide default) policy setting before I disabled third party apps in the Org-wide app setting. The code is only looking at this setting's value now and I think this could be the problem.
Teams third party apps bug 2

See the incorrect policy result below
image

nanda-katikaneni:
#  #
Teams baseline changes based on the comment matrix from pilots and internal test-based feedback.

## 🗣 Description ##
Main changes include:
Reformat the baseline document with new structure and policy identifiers
Reorder couple policies (2 and 3) 
Refine and update policy statements (to remove any ambiguity) based on pilot feedback. 
Update implementation steps for many of policies 

closes #241 
closes #260 
closes #267 
closes #268 
closes #279 
closes #363 
closes #383

## 🧪 Testing ##
Walked through Teams admin center for each policy to confirm implementation steps
Cross-check comment matrix to ensure comments were addressed
Review the changes with another team member.

      in code comments.
- [ ] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Create a release.

### DIFF
--- a/baselines/teams.md
+++ b/baselines/teams.md
@@ -376,9 +376,9 @@ Blog](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/skype-for-busi
 
 ### Implementation
 
-Instructions for *enabling* communications with Skype users are outlined
-in [Communicate with Skype
-users](https://docs.microsoft.com/en-us/microsoftteams/manage-external-access#communicate-with-skype-users).
+All of the settings in this section are configured in the **Microsoft Teams admin center**.
+
+#### MS.TEAMS.6.1v1, instructions:
 
 1.  Sign in to the **[Microsoft Teams admin
     center](https://admin.teams.microsoft.com).**
@@ -418,7 +418,9 @@ available in GCC or DoD tenants.
 
 ### Implementation
 
-To ensure that teams email integration is disabled:
+All of the settings in this section are configured in the **Microsoft Teams admin center**.
+
+#### MS.TEAMS.7.1v1, instructions:
 
 1.  Sign in to the **[Microsoft Teams admin
     center](https://admin.teams.microsoft.com).**
@@ -473,7 +475,9 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/deploy-a
 
 ### Implementation
 
-To restrict which Team apps can be installed:
+All of the settings in this section are configured in the **Microsoft Teams admin center**.
+
+#### MS.TEAMS.8.1v1, instructions:
 
 1.  Sign in to the **[Microsoft Teams admin
     center](https://admin.teams.microsoft.com).**
@@ -482,22 +486,32 @@ To restrict which Team apps can be installed:
 
 3.  Select **Global (Org-wide default)**.
 
-4.  Under **Microsoft apps**, select **Allow all apps**, unless specific
-    apps need to be disallowed, in which case select **Block specific
-    apps and allow all others**.
+4.  Under **Microsoft apps**, select **Allow all apps**
 
-5.  Set **Third-party apps** to **Block all apps**, unless specific apps
+5.  If custom policies have been created, repeat these steps for each
+    policy, selecting the appropriate policy in step 4.
+
+#### MS.TEAMS.8.2v1, instructions:
+
+1.  Sign in to the **[Microsoft Teams admin
+    center](https://admin.teams.microsoft.com).**
+
+2.  Select **Teams apps -**\> **Permission policies.**
+
+3.  Select **Global (Org-wide default)**.
+
+4.  Set **Third-party apps** to **Block all apps**, unless specific apps
     have been approved by the agency, in which case select **Allow
     specific apps and block all others**.
 
-6.  Set **Custom apps** to **Block all apps**, unless specific apps have
+5.  Set **Custom apps** to **Block all apps**, unless specific apps have
     been approved by the agency, in which case select **Allow specific
     apps and block all others**.
 
-7.  Click **Save**.
+6.  Click **Save**.
 
-8.  If custom policies have been created, repeat these steps for each
-    policy, selecting the appropriate policy in step 3.
+7.  If custom policies have been created, repeat these steps for each
+    policy, selecting the appropriate policy in step 4 & 5.
 
 ## 9. Cloud Recording of Teams Meetings
 
@@ -533,7 +547,9 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/policy-assignment-overview
 
 ### Implementation
 
-To configure the Meeting policies for cloud video recording:
+All of the settings in this section are configured in the **Microsoft Teams admin center**.
+
+#### MS.TEAMS.9.1v1, instructions:
 
 1.  Sign in to the **[Microsoft Teams admin
     center](https://admin.teams.microsoft.com).**
@@ -576,6 +592,10 @@ record meetings:
 
 9.  Select **Apply**.
 
+#### MS.TEAMS.9.2v1, instructions:
+
+N/A
+
 ## 10. Recording of Live Events
 
 Live events are recorded by default. Agencies should increase their
@@ -598,6 +618,10 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/teams-live-events/live-eve
 - N/A
 
 ### Implementation
+
+All of the settings in this section are configured in the **Microsoft Teams admin center**.
+
+#### MS.TEAMS.10.1v1, instructions:
 
 1.  Sign in to the **[Microsoft Teams admin
     center](https://admin.teams.microsoft.com).**
--- a/baselines/teams.md
+++ b/baselines/teams.md
@@ -513,6 +513,10 @@ All of the settings in this section are configured in the **Microsoft Teams admi
 7.  If custom policies have been created, repeat these steps for each
     policy, selecting the appropriate policy in step 4 & 5.
 
+#### MS.TEAMS.8.3v1, instructions:
+
+N/A
+
 ## 9. Cloud Recording of Teams Meetings
 
 This setting determines whether video can be recorded in meetings hosted
@@ -579,22 +583,29 @@ record meetings:
 4.  Under the **Recording & transcription** section, set **Cloud
     recording** to **On**.
 
-5.  Under the **Recording & transcription** section, set **Store
-    recordings outside of your country or region** to **Off**.
+5.  Select **Save**.
 
-6.  Select **Save**.
-
-7.  After selecting **Save**, a table displays the set of policies.
+6.  After selecting **Save**, a table displays the set of policies.
     Select the row containing the new policy, then select **Manage
     users**.
 
-8.  Assign the users that need the ability to record to this policy.
+7.  Assign the users that need the ability to record to this policy.
 
-9.  Select **Apply**.
+8.  Select **Apply**.
 
 #### MS.TEAMS.9.2v1, instructions:
 
-N/A
+1.  Sign in to the **[Microsoft Teams admin
+    center](https://admin.teams.microsoft.com).**
+
+2.  Select **Meetings** -\> **Meeting policies**.
+
+3.  Select the **Global (Org-wide default)** policy.
+   
+4.  Under the **Recording & transcription** section, set **Store
+    recordings outside of your country or region** to **Off**.
+
+5.  Select **Save**.
 
 ## 10. Recording of Live Events
 
--- a/baselines/teams.md
+++ b/baselines/teams.md
@@ -153,7 +153,7 @@ All the settings in this section are configured in the [**Microsoft Teams admin
 3.  Select the **Global (Org-wide default)** policy.
 
 4.  Under the **Meeting join & lobby** section **,** ensure
-    **Who can bypass the lobby** is *not* set to **Everyone**.
+    **Who can bypass the lobby** is *not* set to **Everyone**. Bypassing the lobby should be set to **People in my org**, though other options may be used if needed. 
 
 5.  In the same section, set **People dialing in can bypass the lobby** to
     **Off**.
--- a/baselines/teams.md
+++ b/baselines/teams.md
@@ -451,13 +451,13 @@ development, that users “sideload” into Teams.
 Agencies SHOULD "Allow specifc apps and block all others" for Microsoft Apps in the Global (org-wide default) policy, all custom policies, and the Org-wide app settings.
 - _Rationale:_ The security risk with allowing integration for all Microsoft Apps is access to Teams information through vulnerabilties that can be present in app unmanged by the agency. By only allowing specific apps and blocking all others the agency will better manage its app integration and potential exposure points.
 - _Last modified:_ July 2023
-- _Note:_ Custom policies MAY be created to allow more flexibility for specific users.
-- 
+- _Note:_ This policy applies to the Global (Org-wide default) meeting policy and in custom meeting policies if they exist. Custom policies MAY be created to allow more flexibility for specific users.
+ 
 #### MS.TEAMS.8.2v1
 Agencies SHOULD NOT allow installation of all third-party apps and custom apps in the Global (org-wide default) policy, all custom policies, and the Org-wide app settings.
 - _Rationale:_ The security risk with allowing integration third-party and custom is access to Teams information through vulnerabilties that can be present in an app unmanged by the agency. By blocking all third-party apps and custom apps the agency will restrict access of Teams information to external apps. 
 - _Last modified:_ July 2023
-- _Note:_ Custom policies MAY be created to allow more flexibility for specific users. 
+- _Note:_ This policy applies to the Global (Org-wide default) meeting policy and in custom meeting policies if they exist. Custom policies MAY be created to allow more flexibility for specific users.
 
 ### Resources
 
@@ -541,7 +541,7 @@ This section helps reduce security risks related to the recording user permissio
 Meeting recording SHOULD be disabled in the Global (org-wide default) meeting policy.
 - _Rationale:_ The security risk with allowing all users to record Teams meetings and group calls exposes the ability to capture all audio, video, screen sharing activity, and information shared. By disabling the meeting recording setting in the Global (org-wide default) meeting policy an agency will limit this exposure of information. 
 - _Last modified:_ July 2023
-- _Note:_ Custom policies MAY be created to allow more flexibility for specific users.
+- _Note:_ This policy applies to the Global (Org-wide default) meeting policy and in custom meeting policies if they exist. Custom policies MAY be created to allow more flexibility for specific users.
 
 ### Resources
 
--- a/baselines/teams.md
+++ b/baselines/teams.md
@@ -552,79 +552,109 @@ Any product that meets the requirements outlined in this baseline policy may be 
 #### MS.TEAMS.6.2v1 instructions:
 Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation-3) for additional guidance.
 
-## 7. Attachment Scanning
+## 7. Malware
 
-Though any product that fills the requirements outlined in this baseline
-control may be used, for guidance on implementing malware scanning using
-Microsoft Defender, see the “Data Loss Prevention SHALL Be Enabled”
-section of the *Defender for Office 365 Minimum Viable Secure
-Configuration Baseline*.
+Any product that meets the requirements outlined in this baseline
+policy group may be used. If the agency is using Microsoft Defender to
+implement malware scanning, see the following policies of the CISA M365 Security Configuration Baseline for Defender for Office 365 for additional guidance.
+
+- [MS.DEFENDER.1.3v1 \| CISA M365 Security Configuration Baseline for Defender for Office 365](./defender.md#msdefender13v1)
+  - All users SHALL be added to Defender for Office 365 Protection in either the standard or strict preset security policy.
 
 ### Policies
 
 #### MS.TEAMS.7.1v1
 Attachments included with Teams messages SHOULD be scanned for malware.
-- _Rationale:_ TODO
+- _Rationale:_ Teams can be used as a mechanism for delivering malware. In many cases, malware can be detected through scanning, reducing the risk for end users.
 - _Last modified:_ July 2023
 
 #### MS.TEAMS.7.2v1
 Users SHOULD be prevented from opening or downloading files detected as malware.
-- _Rationale:_ TODO
+- _Rationale:_ Teams can be used as a mechanism for delivering malware. In many cases, malware can be detected through scanning, reducing the risk for end users.
 - _Last modified:_ July 2023
 
 ### Resources
 
-- The “Data Loss Prevention SHALL Be Enabled” section of the *Defender for
-Office 365 Minimum Viable Secure Configuration Baseline.*
+- [Recommended settings for EOP and Microsoft Defender for Office 365
+  security \| Microsoft
+  Docs](https://learn.microsoft.com/en-us/microsoft-365/security/office-365-security/recommended-settings-for-eop-and-office365)
+
+### License Requirements
+
+- N/A
+
+### Implementation
+#### MS.TEAMS.7.1v1 instructions:
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation) for additional guidance.
+
+#### MS.TEAMS.7.2v1 instructions:
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation) for additional guidance.
 
 ## 8. Link Protection
 
-Microsoft Defender protects users from malicious links included in Teams
-messages by prepending
+Several technologies exist for protecting users from malicious links
+included in emails. For example, Microsoft Defender accomplishes this by
+prepending
 
-`https://\*.safelinks.protection.outlook.com/?url=`
+`https://*.safelinks.protection.outlook.com/?url=`
 
-to URLs included in the messages. By prepending the safe links URL,
+to any URLs included in emails. By prepending the safe links URL,
 Microsoft can proxy the initial URL through their scanning service.
-Their proxy performs the following checks:
+Their proxy can perform the following actions:
 
-- Compares the URL with a block list
+- Compare the URL with a block list,
 
-- Compares the URL with a list of know malicious sites
+- Compare the URL with a list of know malicious sites, and
 
-- If the URL points to a downloadable file, applies real-time file
-scanning
+- If the URL points to a downloadable file, apply real-time file
+  scanning.
 
-- If all checks pass, the user is redirected to the original URL.
+If all checks pass, the user is redirected to the original URL.
 
-Though Defender’s use is not strictly required for this purpose,
-guidance for enabling link scanning using Microsoft Defender is included
-in the “Safe Links Policies SHALL Be Enabled” and “Safe Links in Global
-Settings SHALL be Configured” sections of the *Defender for Office 365
-Minimum Viable Secure Configuration Baseline.*
+Microsoft Defender includes link scanning capabilities
+Use of Microsoft Defender is not strictly required for this purpose;
+any product that fulfills the requirements outlined in this baseline policy group may be used.
+If the agency is using Microsoft Defender to meet this baseline policy group, see the following policy of the CISA M365 Security Configuration Baseline for Defender for Office 365 for additional guidance.
+
+- [MS.DEFENDER.1.3v1 \| CISA M365 Security Configuration Baseline for Defender for Office 365](./defender.md#msdefender13v1).
+  - All users SHALL be added to Defender for Office 365 Protection in either the standard or strict preset security policy.
 
 ### Policies
 
 #### MS.TEAMS.8.1v1
 URL comparison with a block-list SHOULD be enabled.
-- _Rationale:_ TODO
+- _Rationale:_ Users may be directed to malicious websites via links in Teams. Blocking access to known, malicious URLs can prevent users from accessing known malicious websites.
 - _Last modified:_ July 2023
 
 #### MS.TEAMS.8.2v1
 Direct download links SHOULD be scanned for malware.
-- _Rationale:_ TODO
+- _Rationale:_ URLs in Teams may direct users to download and run malware. Scanning direct download links in real-time for known malware and blocking access can prevent users from infecting their devices.
 - _Last modified:_ July 2023
 
 #### MS.TEAMS.8.3v1
 User click tracking SHOULD be enabled.
-- _Rationale:_ TODO
+- _Rationale:_ Users may click on malicious links in Teams, leading to compromise or authorized data disclosure. Enabling user click tracking lets agencies know if a malicious link may have been visited after the fact to help tailor a response to a potential incident.
 - _Last modified:_ July 2023
 
 ### Resources
 
-- The “Safe Links Policies SHALL Be Enabled” section of the *Defender for
-Office 365 Minimum Viable Secure Configuration Baseline.*
+- [Recommended settings for EOP and Microsoft Defender for Office 365
+  security \| Microsoft
+  Docs](https://learn.microsoft.com/en-us/microsoft-365/security/office-365-security/recommended-settings-for-eop-and-office365)
 
+### License Requirements
+
+- N/A
+
+### Implementation
+#### MS.TEAMS.8.1v1 instructions:
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation) for additional guidance.
+
+#### MS.TEAMS.8.2v1 instructions:
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation) for additional guidance.
+
+#### MS.TEAMS.8.3v1 instructions:
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation) for additional guidance.
 
 # Acknowledgements
 
--- a/baselines/teams.md
+++ b/baselines/teams.md
@@ -76,7 +76,7 @@ External meeting participants SHOULD NOT be enabled to request control of shared
 ### Resources
 
 - [Manage who can present and request control in Microsoft Teams \| Microsoft
-  Docs] (https://learn.microsoft.com/en-us/microsoftteams/meeting-who-present-request-control) 
+  Docs](https://learn.microsoft.com/en-us/microsoftteams/meeting-who-present-request-control) 
   
 
 ### License Requirements
@@ -131,7 +131,7 @@ B2B guest users MAY be admitted automatically in the Global (Org-wide default) m
 
 ### Resources
 - [Meeting policy settings \| Microsoft
-  Docs] (https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#meetings) 
+  Docs](https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#meetings) 
   
 ### License Requirements
 - N/A
@@ -199,7 +199,7 @@ Anonymous users SHALL NOT be enabled to start meetings.
 
 ### Resources
 - [Meeting policy settings  \| Microsoft
-Docs] (https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#meetings) 
+Docs](https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#meetings) 
 
 ### License Requirements
 - N/A
@@ -374,7 +374,7 @@ Contact with Skype users SHALL be blocked.
 ### Resources
 
 - [Communicate with Skype users \| Microsoft
-Docs] (https://learn.microsoft.com/en-us/microsoftteams/external-meetings-skype-for-business-server-hybrid)
+Docs](https://learn.microsoft.com/en-us/microsoftteams/external-meetings-skype-for-business-server-hybrid)
 
 - [Skype for Business Online to Be Retired in 2021 \| Microsoft Teams
 Blog](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/skype-for-business-online-to-be-retired-in-2021/ba-p/777833)
@@ -412,7 +412,7 @@ Teams email integration SHALL be disabled.
 ### Resources
 
 - [Email Integration \| Microsoft
-Docs] (https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#email-integration)
+Docs](https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#email-integration)
 
 ### License Requirements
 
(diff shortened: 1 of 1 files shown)


baselines/teams.md
```diff
@@ -339,34 +388,27 @@ Blog](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/skype-for-busi
 
 ### Implementation
 
-Instructions for *enabling* communications with Skype users are outlined
-in [Communicate with Skype
-users](https://docs.microsoft.com/en-us/microsoftteams/manage-external-access#communicate-with-skype-users).
+All the settings in this section are configured in the **Microsoft Teams admin center**.
```
adhilto, line 391: For similar statements made for MS.TEAMS.1 through MS.TEAMS.5, "Microsoft Teams admin center" was a link. Recommend changing this to a link for consistency.

baselines/teams.md
```diff
@@ -381,19 +423,20 @@ available in GCC or DoD tenants.
 
 ### Implementation
 
-To ensure that teams email integration is disabled:
+All the settings in this section are configured in the **Microsoft Teams admin center**.
```
adhilto, line 426: For similar statements made for MS.TEAMS.1 through MS.TEAMS.5, "Microsoft Teams admin center" was a link. Recommend changing this to a link for consistency.

baselines/teams.md
```diff
@@ -434,53 +473,59 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/deploy-a
...
+- _Note:_ Custom policies MAY be created to allow more flexibility for specific users.
 
 #### MS.TEAMS.9.2v1
- Alternate meeting policies MAY be created that allow agency-approved users the ability to record.
-- _Rationale:_ TODO
-- _Last modified:_ July 2023
-
-#### MS.TEAMS.9.3v1
-For all meeting polices that allow cloud recording, recordings SHOULD be stored inside the country of that agencys tenant.
+For all meeting polices (Global (org-wide default) meeting policy and all custom policies) that allow cloud recording, recordings SHOULD be stored inside the country of that agencys tenant.
```
adhilto, line 528: Minor typo
```suggestion
For all meeting polices (Global (org-wide default) meeting policy and all custom policies) that allow cloud recording, recordings SHOULD be stored inside the country of that agency's tenant.
```

baselines/teams.md
```diff
@@ -498,12 +543,14 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/policy-assignment-overview
 
 ### Implementation
 
-To configure the Meeting policies for cloud video recording:
+All the settings in this section are configured in the **Microsoft Teams admin center**.
```
adhilto, line 546: For similar statements made for MS.TEAMS.1 through MS.TEAMS.5, "Microsoft Teams admin center" was a link. Recommend changing this to a link for consistency.

baselines/teams.md
```diff
@@ -512,45 +559,55 @@ To configure the Meeting policies for cloud video recording:
...
+
+## 10. Recording of Live Events
+This section helps reduce security risks related to the recording user permissions for Live Events.
 
 ### Policies
 #### MS.TEAMS.10.1v1
-Record an event SHOULD be set to Organizer can record.
-- _Rationale:_ TODO
+Record an event SHOULD be set to Organizer can record Global (org-wide default) meeting policy and all custom policies.
+- _Rationale:_ The security risk of the default settings for Live Events are that Live Events are avaialble to be recoreded by all participants by default. By limiting the recording permissions to only the organizer this minimizes the security risk to the organizer's discretion for these Live Events.
```
adhilto, line 608: Minor typos
```suggestion
- _Rationale:_ The security risk of the default settings for Live Events is that Live Events are available to be recorded by all participants by default. By limiting the recording permissions to only the organizer this minimizes the security risk to the organizer's discretion for these Live Events.
```

Also, I get what we're saying here, but the justification here feels weak to me. What is the security risk of having a live event recorded? Assuming proper SharePoint controls are in place, the recordings should be safe. Plus these live events things tend to be intended for large audiences, such as a town hall meeting. Meaning you wouldn't expect them to have info that isn't appropriate for the org at-large in the first place. Have we considered cutting this control? I think I remember some agencies expressing similar feelings during the pilot meetings, but not entirely sure.

baselines/teams.md
```diff
@@ -563,17 +620,24 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/teams-live-events/live-eve
 
 ### Implementation
 
+All the settings in this section are configured in the **Microsoft Teams admin center**.
```
adhilto, line 623: For similar statements made for MS.TEAMS.1 through MS.TEAMS.5, "Microsoft Teams admin center" was a link. Recommend changing this to a link for consistency.

baselines/teams.md
```diff
@@ -563,17 +620,24 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/teams-live-events/live-eve
...
 3.  Select **Global (Org-wide default)**.
 
 4.  Set **Record an event** to **Organizer can record**.
 
 5.  Click **Save**.
 
+6.  If custom policies have been created, repeat these steps for each
+    policy, selecting the appropriate policy in step 3.
+
 ## 11. Data Loss Prevention
```
adhilto, line 641: Looks like MS.TEAMS.11.2v1 and MS.TEAMS.11.3v1 are exact copies of each other.

Also, it may still be too soon to change these, but do we have an issue to update the resource sections that point to Defender with a more precise pointer? I do think this change should be a part of Emerald.

For example, change
> The “Data Loss Prevention SHALL Be Enabled” section of the Defender for Office 365 Minimum Viable Secure Configuration Baseline.

to 
> [Section 4, Data Loss Prevention | Defender for Office 365 Minimum Viable Secure Configuration Baseline](https://github.com/cisagov/ScubaGear/blob/260-revise-teams-policy-24-based-on-the-feedback-and-internal-testing/baselines/defender.md#4-data-loss-prevention).

adhilto (changes requested):

baselines/teams.md
```diff
@@ -88,114 +87,141 @@ need.
 - N/A
 
 ### Implementation
-
+All the settings in this section are configured in the [**Microsoft Teams admin
+    center**](https://admin.teams.microsoft.com).
+#### MS.TEAMS.1.1v1, instructions:
```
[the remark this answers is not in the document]
tkol2022: Let me know what you all decide on this. No preference from me.
If we change it, we need to update the [markdown style guide](https://github.com/cisagov/ScubaGear/blob/Baseline-Markdown-Style-Guide/baselines/Example.md) and at a minimum AAD and SharepointOneDrive because I'm using the same format as this Teams pull request branch.

baselines/teams.md
```diff
@@ -88,114 +87,141 @@ need.
...
+Anonymous users, including dial-in users, SHOULD NOT be admitted automatically in the Global (Org-wide default) meeting policy.
 - _Rationale:_ TODO add rationale.
 - _Last modified:_ July 2023
+- Custom meeting policies MAY be created that allow more flexibility for specific users.
+
+#### MS.TEAMS.2.2v1
+Internal users SHOULD be admitted automatically in the Global (Org-wide default) meeting policy.
+- _Rationale:_ TODO add rationale.
+- _Last modified:_ July 2023
+- Custom meeting policies MAY be created that allow more flexibility for specific users.
```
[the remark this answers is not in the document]
tkol2022: Agree with @adhilto - place this in a Note field.
Question: What is the purpose of this sentence about custom meeting policies? If we removed it, what would be the impact to the user?
2nd question: Can a customer create a custom meeting policy that overrides the Global policy? If yes, then maybe the note should be a warning to the user that a custom policy can affect the security of the tenant even if a global policy is set.

tkol2022:
@nanda-katikaneni @ahuynhMITRE I noticed that in the instructions for numerous items we ask the user the repeat a set of steps for custom policies. We also mention that custom policies allow for flexibility and and can be leveraged for specific users or use cases.

- Do custom policies override the Global (Org-wide default)?
- It is not clear what our stance, or more specifically our instructions are, regarding custom policies? Are we telling agencies that they should use them and then if they do, simply configure them just like we instruct to configure the Global (Org-wide default) policy? By using the word "custom", it can imply that a policy is different from the Global (Org-wide default) and does this mean that custom policies can be configured to a less secure state? For me, there is ambiguity on what we expect from users regarding custom policies and their configuration.

tkol2022:
Sections 1,2,3,9 are all related to configurations on meeting policies and seem to be configured from the same admin page. Can we logically combine these into a section named Meeting Policies? It is not obvious why they should be logically separated.

tkol2022:
This comment applies to all Rationale sections. Remove any language that quantifies the risk of not implementing the secure baseline policy. For example, "Furthermore, **the risk is minimal as users cannot gain control** of another user’s screen unless the user giving control explicitly accepts a control request." Let's stay away from saying "the risk is minimal", etc. unless there is a strong case for doing so. In this example rationale, I don't think there is.

We could rephrase this Rationale as follows:

`There is risk in granting an external participant control of a shared screen because the participant could potentially perform unauthorized actions on the shared screen. This policy reduces that risk by removing the ability of an external participant to request control. `

Finally, if there are legitimate use cases from the agency pilots for allowing external participants to have shared control, please describe them here for awareness purposes.

tkol2022:
Perform a review of all the links in your Resources sections and make sure that the configuration guidance/instructions described in those links matches with the specific configurations in the respective baseline policy.
For example, section 1 links to a page that doesn't seem to contain the relevant configuration. I found [this page instead](https://learn.microsoft.com/en-us/microsoftteams/meeting-who-present-request-control).

baselines/teams.md
ahuynhMITRE: I did not change the content in the last three policies since they are directed at defender. I agree with holding off until defender updates are in to have a more precise pointer and use the same language throughout the policy group and policies.

baselines/teams.md
```diff
@@ -55,144 +56,175 @@ document. The United States Government has adapted selections of these
...
+4.  Under the **Meeting join & lobby** section **,** ensure
+    **Who can bypass the lobby** is set to **People in my org**.
 
-5.  If custom policies have been created, repeat these steps for each
-    policy, selecting the appropriate policy in step 3.
+5.  In the same section, set **People dialing in can bypass the lobby** to
+    **Off**.
 
-## 2.3 Automatic Admittance to Meetings SHOULD Be Restricted
+#### MS.TEAMS.2.3v1, instructions:
```
tkol2022, line 175: After reviewing the instructions for MS.TEAMS.2.1v1, MS.TEAMS.2.2v1, and MS.TEAMS.2.3v1, they seem to be the same policy. Even though the policy statements are worded differently and mention different target user groups, the implementation makes it clear that all of these are referencing the "Who can bypass the lobby" setting under the Meeting Policies > Global (Org-wide default) policy.

My recommendation is to combine these three into a single policy which may read something like "****Everyone** SHOULD NOT be admitted automatically in the Global (Org-wide default) meeting policy**" or similar. The term Everyone in this context refers to the setting value of "Everyone" in the drop down box. Any value other than Everyone would be compliant with the policy. There are numerous compliant values that an agency user could select including People in my org and guests, People in my org, trusted orgs and guests, People in my org, and more. ScubaGear would produce a Pass for any of those values, as long as Everyone is not the selected value.
I think this will give flexibility to agencies and they can decide how strict they want to make the setting (as long as they don't allow Everyone to automatically join) for their use cases.

image

adhilto: I get that "Everyone" is an option from the dropdown, but I think "Everyone SHOULD NOT be admitted automatically" is confusing. To me it reads as "No one should be admitted automatically," which isn't what we're going for at all.

Also, "just don't select everyone" isn't actually equivalent to our current policies, as "only organizers and co-organizers" and "people who were invited" aren't options that meet the current MS.TEAMS.2.2v1. I think the rationale for dictating that internal users should be admitted automatically is to reduce "admit fatigue," if users have to manually admit everyone they might be more likely to admit everyone without thought. You could argue that this isn't a big deal, and I could accept that, but I wanted to at least point it out.

baselines/teams.md
```diff
@@ -88,114 +87,141 @@ need.
...
     center**](https://admin.teams.microsoft.com).
+2.  Select **Meetings >** **Meeting** **policies**.
+
+3.  Select the **Global (Org-wide default)** policy.
 
-2.  Select **Meetings -\>** **Meeting policies**.
+4.  Under the **Meeting join & lobby** section **,** ensure
+    **Who can bypass the lobby** is *not* set to **Everyone**. Bypassing the lobby should be set to **People in my org**, though other options may be used if needed. 
+
+5.  In the same section, set **People dialing in can bypass the lobby** to
```
tkol2022, line 158: I think the configuration for "People dialing in can bypass the lobby SHOULD be set to off" should be represented in the policy description because it is a separate config in the Teams admin center. Instead of trying to add it to the existing policy, maybe just create a new one in this same section. It is currently missing from the policies. We had a similar case in AAD and I ended up creating a policy for it during Emerald.

image

tkol2022:
### General comment
We have numerous policy names that contain the text "Global (Org-wide default) meeting policy or in custom meeting policies if any exist" and slight variations of that. I believe that this is an implementation detail and we might be able to make the policy titles more readable if we remove that part and then describe it in the Notes section. Here is an example for a revised policy.

#### MS.TEAMS.3.1v1
Anonymous users SHALL NOT be enabled to start meetings.

- _Rationale:_ For any agencies that implemented custom policies that provided more flexibility to some users to automatically admit "everyone" to a meeting - this policy provides protection from anonymous users starting meeting to scrape internal contacts. 
- _Last modified:_ July 2023
- _Note:_ This policy applies to the Global (Org-wide default) meeting policy and in custom meeting policies if they exist.

baselines/teams.md
```diff
@@ -88,114 +87,141 @@ need.
...
-
-5.  In the same section, set **Dial-in users can bypass the lobby** to
-    **Off**.
+4.  Under the **Meeting join & lobby** section **,** set **Anonymous users and dial-in callers can start a meeting** to **Off.**
 
-6.  If custom policies have been created, repeat these steps for each
+5.  If custom policies have been created, repeat these steps for each
     policy, selecting the appropriate policy in step 3.
 
 ## 4. External User Access
```
tkol2022, line 227: The hyperlink to the OMB Max Site is presumably there to provide a list of .gov domains but what is a user supposed to do with this information?

- The OMB Max Site requires a login. Are all government users supposed to be able to login there? Once they are in, what are they supposed to do? We don't provide any instructions for navigating that site. 
- Are we expecting a user to enter the list of every government .gov domain into the Teams allowable domain configuration?

baselines/teams.md
```diff
@@ -1,11 +1,12 @@
-# Introduction
+# CISA M365 Security Configuration Baseline for Teams
 
 Microsoft Teams is a text and live chat workspace in Microsoft 365 that
 supports video calls, chat messaging, screen-sharing, and file sharing.
 It has a permission-based team structure for managing calls and files.
 Microsoft teams also enables teams to manage their own user access
 rights, security policies, and record video calls.
 
+## Key Terminology
```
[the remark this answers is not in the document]
nanda-katikaneni: Good catch - addressed in the latest push.

baselines/teams.md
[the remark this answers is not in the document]
nanda-katikaneni: Based on latest style guide changes, removed the comma in all instructions sub-headings.

baselines/teams.md
nanda-katikaneni: Updated to add a link.

baselines/teams.md
nanda-katikaneni: Updated to add a link.

baselines/teams.md
nanda-katikaneni: Updated to add a link.

baselines/teams.md
nanda-katikaneni: Updated to add a link.

baselines/teams.md
```diff
@@ -39,7 +40,7 @@ or [G3](https://www.microsoft.com/en-us/microsoft-365/government)
 license level. Therefore, only licenses not included in E3/G3 are
 listed.
 
-## 1.2 Resources
+## Resources
```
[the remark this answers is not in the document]
nanda-katikaneni: Updated sections to the new style guide format.

baselines/teams.md
```diff
@@ -55,144 +56,175 @@ document. The United States Government has adapted selections of these
...
-# 2. Baseline
+# Baseline
 
-## 2.1 External Participants SHOULD NOT Be Enabled to Request Control of Shared Desktops or Windows in Meetings
+## 1. Requesting Control of Shared Desktops or Windows
 
 This setting controls whether external meeting participants can request
 control of the shared desktop or window during the meeting. In this
 instance, the term “external participants” includes external users, B2B
 guest users, unmanaged users and anonymous users.
```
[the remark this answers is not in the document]
nanda-katikaneni: Updated all section headings to the example style (except for sections 11,12,13 which are defender dependent and need discussion).

baselines/teams.md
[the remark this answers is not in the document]
nanda-katikaneni: Updated to add a Note. 
About the custom policies: they do override global policy. Custom policies are assigned to specific users or groups to provide more flexibility. Policy 2.x need further fine-tuning (tbd)

baselines/teams.md
nanda-katikaneni: Policy 2.1 includes dialing in users:
MS_TEAMS_2 1

baselines/teams.md
nanda-katikaneni: IMO this group still need three policies may be with modified Notes. 
The first sub-policy that block automatic admittance for anonymous users may not need a custom policy exemption. The second one to allow internal users is to avoid admit fatigue as Alden called above. The third "MAY" policy need better clarification in the notes about custom policy aspect. We can discuss this along with a possible grouping of 1,2,3 and 9.

tkol2022:
I just tested the hyperlinks to the Teams admin center in your instructions sections and they do not seem to work with our GCC High test tenant. See the error message screenshot below. The Teams link for GCC High is https://admin.gov.teams.microsoft.us/ but I'm not sure it makes sense to include links in our documentation. Let's discuss at tomorrow's stand-up.

image

baselines/teams.md
```diff
@@ -295,40 +320,61 @@ Docs](https://docs.microsoft.com/en-us/microsoftteams/manage-external-access#man
...
+    
 4.  To allow contact with unmanaged users only if the internal user
     initiates the contact:
 
     1.  Under **Teams accounts not managed by an organization**, set **People in my organization can communicate with Teams users whose accounts aren't managed by an organization** to **On**.
 
     2.  Clear the check next to **External users with Teams accounts not managed by an organization can contact users in my organization**.
 
+
+#### MS.TEAMS.5.2v1 instructions:
```
tkol2022, line 344: The policies in Section 5 are confusing and appear to conflict with each other. Let's discuss these verbally. Policy 5.1 seems to have two different compliant configuration states (as noted by instructions steps 3 and 4 which cannot be configured at the same time since they represent two distinct states. Policy 5.2 has a single compliant configuration state (noted by instruction step 3) which is identical to policy 5.1 step 3 and therefore seems duplicative.

Overall, I cannot think of a good business case where you would want a federal agency user to communicate with an unmanaged user (i.e. does not belong to an organization on the approved domain list from policy MS.TEAMS.4.1v1). We can discuss so I can hear your feedback from the pilots and ideas.  

I believe there should only be a single compliant policy / configuration: Completely block contact with unmanaged users: under **Teams accounts not managed by an organization**, set **People in my organization can communicate with Teams users whose accounts aren't managed by an organization** to **Off**.

tkol2022:
Recommend to group policies 4, 5 and 6 into a section named External Access Restrictions or something like that.
All of the policies in this section are configured in a Teams admin screen named **External Access** and are logically related.
I've found that the baseline documents are much easier to read when more policies are grouped. AAD had a large portion of the policies grouped into related sections.

image

baselines/teams.md
```diff
@@ -339,40 +385,34 @@ Blog](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/skype-for-busi
...
-agencies do not have control over the security settings associated with
-this email. For this reason, email channel integration should be
-disabled.
+This section helps reduce security risks related to teams email integration. Teams provides an optional feature that allows channels to have an email address and receive email.
 
 ### Policies
 #### MS.TEAMS.7.1v1
 Teams email integration SHALL be disabled.
-- _Rationale:_ TODO
+- _Rationale:_ The security risk with Microsoft Team's email integration is due to the fact that Teams channel's email addresses are not under the tenant’s domain, rather they are associated with the Microsoft-owned domain of "teams.ms". By disabling Teams email integration the risk of use of emails outside of the agency's security settings control is minimized.
```
tkol2022, line 409: I do not understand what is meant by "By disabling Teams email integration the risk of use of emails outside of the agency's security settings control is minimized."
Needs clarity.

baselines/teams.md
```diff
@@ -339,40 +385,34 @@ Blog](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/skype-for-busi
...
 - _Last modified:_ July 2023
 
 ### Resources
 
 - [Email Integration \| Microsoft
-Docs](https://docs.microsoft.com/en-us/microsoftteams/enable-features-office-365#email-integration)
+Docs](https://learn.microsoft.com/en-us/microsoftteams/settings-policies-reference#email-integration)
 
 ### License Requirements
```
tkol2022, line 417: The requirements state that teams email integration is not available in certain tenants. Does the ScubaGear code take this into account and not produce a false negative?

baselines/teams.md
```diff
@@ -381,19 +421,21 @@ available in GCC or DoD tenants.
...
+
+#### MS.TEAMS.7.1v1 instructions:
 
 1.  Sign in to the **[Microsoft Teams admin
     center](https://admin.teams.microsoft.com).**
 
-2.  Select **Teams** -\> **Teams Settings**.
+2.  Select **Teams** > **Teams Settings**.
 
 3.  Under the **Email integration** section, set **Allow users to send
```
tkol2022, line 434: The actual configuration setting on the admin center screen is worded slightly different from the instructions.

image

baselines/teams.md
adhilto: Yes. In fact I just fixed a bug with how we were checking the tenant type. https://github.com/cisagov/ScubaGear/pull/333

tkol2022:
@nanda-katikaneni @ahuynhMITRE @adhilto With respect to policy MS.TEAMS.8.2v1 that deals with "third-party apps", I read in the [Microsoft documentation](https://learn.microsoft.com/en-us/microsoftteams/manage-apps#manage-org-wide-app-settings-for-microsoft-365-government) that "In a Microsoft 365 Government - GCC, GCCH and DoD deployment of Teams, **all third-party apps are blocked by default**. In **GCCH and DOD clouds, the third-party apps aren't available**."

I know there was a recent fix 333 related to the tool producing incorrect results for a the Teams email policy in GCC environments.

- Are you sure that the tool produces correct results for MS.TEAMS.8.2v1 in the GCC and DoD environments?
- Based on what Microsoft said above, is policy MS.TEAMS.8.2v1 "not applicable" in the aforementioned environments? In other words it doesn't make sense to check it because there is no way to configure it, or configuring it doesn't make a practical difference because third party apps cannot be installed by users?

adhilto:
> @nanda-katikaneni @ahuynhMITRE @adhilto With respect to policy MS.TEAMS.8.2v1 that deals with "third-party apps", I read in the [Microsoft documentation](https://learn.microsoft.com/en-us/microsoftteams/manage-apps#manage-org-wide-app-settings-for-microsoft-365-government) that "In a Microsoft 365 Government - GCC, GCCH and DoD deployment of Teams, **all third-party apps are blocked by default**. In **GCCH and DOD clouds, the third-party apps aren't available**."
> 
> I know there was a recent fix 333 related to the tool producing incorrect results for a the Teams email policy in GCC environments.
> 
> * Are you sure that the tool produces correct results for MS.TEAMS.8.2v1 in the GCC and DoD environments?
> * Based on what Microsoft said above, is policy MS.TEAMS.8.2v1 "not applicable" in the aforementioned environments? In other words it doesn't make sense to check it because there is no way to configure it, or configuring it doesn't make a practical difference because third party apps cannot be installed by users?

You're right, we did code a special check for the Teams email integration. As far as I'm aware we haven't done that for 8.2. I wasn't aware that third party apps weren't available for GCCH. As for it being "not applicable" in those environments, I vote we make it so that 8.2 automatically passes if we detect that it's a GCCH or DOD tenant (we should create a separate issue for this if we're all in agreement). We should still keep the policy and check for the other environments though, as well as clarify in the license requirements section that third party apps aren't available for GCCH or DOD.

tkol2022:
For policy MS.TEAMS.8.1v1 "Agencies SHOULD allow all apps published by Microsoft", I do not understand how this mitigates a specific security risk?

The other available options for this setting are: Allow specific apps and block all others, Block specific apps and allow all others and Block all apps.

- I would think that Allow specific apps and block all others, is also a reasonable choice, especially since it is "more secure" since it is is more restrictive. Therefore I don't think it would make sense to Fail an agency if they are more restrictive. We had to make similar tweaks to the Sharepoint/OneDrive settings so that we don't produce a Fail when a setting is configured more restrictive.
- The more important question though is, can you  define a clear security risk mitigation for this policy? What is is trying to achieve? For example with third-party apps (which is policy 8.2) if you define the policy to Allow only approved third-party apps, that seems like a clear security benefit.

tkol2022:
> requirements section that third party apps aren't available for GCCH or DOD.

I highly recommend some hands-on testing of the relevant Teams functionality before any decisions are made. 

- See what options are available in the Teams admin center. For example, in a GCC High environment (or GCC), do you even have the option to configure Third-party app restrictions?
- Run the Teams client on a virtual machine with a test user and see what the practical impacts are. For example, regardless of what the options are in the Teams admin center, I would run a test with the Teams client in a GCC environment and see if the user has the option to install 3rd party apps?
- Run the tests in commercial, GCC and GCC High tenants to confirm any differences both in the Teams admin center options and differences in the client user experience.

If you all agree, please open a hands-on exploration issue.

tkol2022:
> > requirements section that third party apps aren't available for GCCH or DOD.
> 
> I highly recommend some hands-on testing of the relevant Teams functionality before any decisions are made.
> 
> * See what options are available in the Teams admin center. For example, in a GCC High environment (or GCC), do you even have the option to configure Third-party app restrictions?
> * Run the Teams client on a virtual machine with a test user and see what the practical impacts are. For example, regardless of what the options are in the Teams admin center, I would run a test with the Teams client in a GCC environment and see if the user has the option to install 3rd party apps?
> * Run the tests in commercial, GCC and GCC High tenants to confirm any differences both in the Teams admin center options and differences in the client user experience.
> 
> If you all agree, please open a hands-on exploration issue.

I forgot to mention that policy MS.TEAMS.8.2v1 also includes "custom apps" not just third-party apps, even though custom apps are not explicitly mentioned in Microsoft's caveat. Therefore the exploration should include custom apps to see the difference between available configuration options across tenant types. Custom Apps is a separate configuration setting from "Third-party apps" in the portal.

tkol2022:
Regarding policy MS.TEAMS.8.2v1 "Agencies SHOULD NOT allow installation of all third-party apps or custom apps in the Global (org-wide default) meeting policy and all custom policies", I found out that there is a separate configuration screen named the **Org-wide app settings** which appears to override the values configured in the **Global (Org-wide default)** policy referenced in the baseline instructions based on my hands-on testing with a Teams client and test user.

This impacts:

- The teams MS.TEAMS.8.2v1 policy statement because it explicitly mentions the Global (org-wide default) meeting policy.
- The policy instructions section. You probably need to include instructions for configuring the **Org-wide app settings** and make other tweaks.
- The ScubaGear code because it evaluates the policy check by only examining the Global (org-wide default) meeting policy.

See existing issue #383 for details and screenshots.

tkol2022:
For MS.TEAMS.9.1v1 instructions, I don't see an option named "**Cloud recording**". Maybe it got renamed to "**Meeting recording**"? 

image

tkol2022:
For MS.TEAMS.9.1v1 instructions, do we need the extra instructions in the section that starts with "If there is a legitimate business need, specific users can be given permission to record meetings"? Those extra instructions are not part of the baseline policy, but seem to be informational to the user, helping educate them on how to tailor policies. Are those steps needed to implement the baseline policy?

image

tkol2022:
Regarding MS.TEAMS.9.2v1 instructions and storing of meeting recordings outside of the country:

- By default [Teams stores meeting recordings in OneDrive or Sharepoint](https://learn.microsoft.com/en-us/microsoftteams/meeting-recording?tabs=meeting-policy). If you turn on the config setting "**Store recordings outside of your country or region**", presumably this gives the user the ability to store the meetings recordings somewhere else (i.e. outside the country), but what does that actually mean in practice? If you turn this on, what actual technical options does a user get when they record a meeting and what is the security risk associated with those options?

The answer to this question can help us write the Rationale for the policy.

tkol2022:
For MS.TEAMS.10.1v1 instructions, step 4, the only compliant configuration option is "**Organizer can record**". Should we also allow the more restrictive setting "**Never record**" in case an agency implements a stricter security policy?

tkol2022:
Policy groups 11, 12 and 13 appear to be policy items that are handled by the Defender baseline. As such I don't see a reason to have them here because of the following reasons:

- The less duplication, the better. It doesn't make sense to me to create an entire section of content that is verbatim the same as a section in the Defender baseline. Either just point to the content in Defender and do not repeat it here, or do not include the content in this Teams baseline at all.
- The way that the policies in sections 11, 12 and 13 are written, they read like a cross between generic policies (that can be implemented by a third-party DLP solution) and policies specific to Microsoft products (namely Defender). This is confusing (to me). I believe the intent of the security baselines is to offer **specific** configuration guidance on how to secure a tenant and in this case MS Teams. If the policies are generic, then the baseline becomes closer to a NIST 800-53 style requirements manual, which to me, is not effective and not consistent with intent of hardening guidance.

adhilto:
> Policy groups 11, 12 and 13 appear to be policy items that are handled by the Defender baseline. As such I don't see a reason to have them here because of the following reasons:
> 
> * The less duplication, the better. It doesn't make sense to me to create an entire section of content that is verbatim the same as a section in the Defender baseline. Either just point to the content in Defender and do not repeat it here, or do not include the content in this Teams baseline at all.
> * The way that the policies in sections 11, 12 and 13 are written, they read like a cross between generic policies (that can be implemented by a third-party DLP solution) and policies specific to Microsoft products (namely Defender). This is confusing (to me). I believe the intent of the security baselines is to offer **specific** configuration guidance on how to secure a tenant and in this case MS Teams. If the policies are generic, then the baseline becomes closer to a NIST 800-53 style requirements manual, which to me, is not effective and not consistent with intent of hardening guidance.

Yes, these items are handled by the Defender baseline. We actually had quite a few conversations about this topic when we initially drafted the baselines. What we ended up with was an imperfect compromise. The problem with just cutting these items from the Teams baseline is that Defender itself is not required:
> Generally, use of Microsoft Defender is not required by the baselines of the core M365 products (Exchange Online, Teams, etc.). However, some of the controls in the core baselines require the use of a dedicated security tool, such as Defender. This baseline should not be a requirement to use Defender, but instead, as guidance for how these requirements could be met using Defender, should an agency elect to use Defender as their tool of choice.

So if we cut out these items from the Teams baseline, then there no longer is any requirement for Teams to have DLP, or attachment scanning, etc. On the other hand, if we cut the content from the Defender baseline and had the full content in the Teams baseline itself, we'd have to duplicate the full Defender content in EXO as well, as those items apply to both baselines. And it would make it much more unclear that those requirements could be met by third-party software.

There's definitely lots of room for improvement here. But unfortunately it's not as simple as cutting the content from Teams/EXO. What did you have in mind with just pointing to the content in Defender? That's more or less what we tried to do: have a lightweight version of the policy in Teams/EXO that point to the Defender baseline for the full content.

Also see https://github.com/cisagov/ScubaGear/issues/401, a related issue.

adhilto (approved):

tkol2022 (approved):
I provided numerous comments and met with Andrew / Nanda to discuss the items. They have incorporated the feedback and initiated discussions with the broader team for specific items that cut across the documents.

Issue #315, opened 2023-05-02
PowerBI Baseline Changes


amart241:
# 💡 Summary #

What is the work, as a high-level summary?

There are a few changes to the admin portal for PowerBI and the implementation section of the baseline will need to reflect be updated

## Motivation and context ##

Why does this work belong in this project?

Updated baselines are a necessity for this to be a useful to tool to agencies and the public.

This would be useful because...

accurate information on how to be in compliance is the foundation of this project.

## Implementation notes ##

Please provide details for implementation, such as:

Go through the PowerBI baseline and try to follow the steps for implementation

## Acceptance criteria ##

How do we know when this work is done?

When the baseline is accurate to the admin portal

- [ ] Criterion

amart241:
Baseline 1. External Sharing, there is no external sharing toggle in the Export and Sharing Settings. The closest toggle to flip is "Allow Azure Active Directory guest users to access Power BI" and it is not the same setting and this toggle is already addressed in Baseline 3. PowerBI Guest Access. 

Baseline 5. External Editing Capability, there is no "Allow Azure Active Directory guest users to edit and manage content in the organization" toggle in the Export and Sharing Settings.

amart241:
According to this [link](https://learn.microsoft.com/en-us/power-bi/admin/service-admin-portal-export-sharing), there is no mention to a setting explicitly called external sharing. I have also not found an alternative anywhere else in the admin portal. 
It is my suggestion that we remove baseline 1.

In that same link I posted above there is reference to the baseline 5 toggle but in the admin portal itself that toggle doesn't exist. This [link](https://learn.microsoft.com/en-us/power-bi/enterprise/service-admin-azure-ad-b2b#admin-info-for-b2b-collaboration) also has no mention of that setting toggle. 
It is my suggestion that we remove baseline 5 as well.

PR #403, opened 2023-06-26
Adjudicate substantive changes to the powerbi baseline markdown following the new format

amart241:
#  #

## 🗣 Description ##

Changed formatting and content of PowerBI baseline to match new formatting structure

Consolidated format and clearer to read and understand layout

## 🧪 Testing ##

N/A

### DIFF
--- a/baselines/powerbi.md
+++ b/baselines/powerbi.md
@@ -102,11 +102,18 @@ The Publish to Web feature SHOULD be disabled unless the agency mission requires
 
 ### Implementation 
 #### MS.POWERBI.1.1v1, instructions:
-Navigate to the **PowerBI Admin Portal**
 
-1. **Tenant Settings**
-2. **Export and sharing settings**
-3. **Publish to web** > **Disabled**
+1. Navigate to the **PowerBI Admin Portal**
+
+2. Click on the 3 dots in the upper right hand corner
+
+3. Click on Settings > Admin Portal
+
+4. Click on **Tenant Settings**
+
+5. Scroll to **Export and sharing settings**
+
+6. Click **Publish to web** set to **Disabled**
 
 ## 2. Power BI Guest Access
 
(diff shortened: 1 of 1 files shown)


baselines/powerbi.md
```diff
@@ -137,9 +83,9 @@ security groups, instead of allowing the entire agency to publish data
 to the web.
 
 ### Policies
-#### MS.POWERBI.2.1v1
-The Publish to Web feature SHOULD be disabled unless the agency mission requires the capability.
-- _Rationale:_ TODO
+#### MS.POWERBI.1.1v1
+Disable the Publish to Web feature unless the agency mission requires the capability.
```
ahuynhMITRE, line 87: "shall" and "should" language is still expected in the policies. Only removed from the previous "policy groups".

ahuynhMITRE: left this comment only for the first policy to not create duplicate comments

ahuynhMITRE (changes requested):
agree with David's comments and main comment from me was re-adding the "shall" and "should" language  into each of the policies.

ahuynhMITRE:
some comments: 

1. Update text following each policy group to the  match the guide below:

This section helps reduce security risks related to [Policy Group]. This includes [Scope of policies ] (This sentence is as needed).
Example: "This section helps reduce security risks related to the sharing of files with users external to the agency. This includes guest users, users who use a verification code and users who access an Anyone link."

2. remove links to admin centers and re-add them as a generic step. This was recently voted on as the current way forward by by the team.

amart241:
I have removed a total of 3 policies from the Power BI baseline. I most recently removed the logging related policy because it is inconsistent across the new "fabric" and current Power BI, it also requires and Azure subscription. #315 were the other 2 policies that I removed.

ahuynhMITRE:
> This comment is tied to the text after each policy group. See policy group 1 below:

"1. Publish to Web
Power BI has a capability to publish reports and content to the web. This capability creates a publicly accessible web URL that does not require authentication or status as an AAD user to view it. While this may be needed for a specific use case or collaboration scenario, it is a best practice to keep this setting off by default to prevent unintended and potentially sensitive data exposure.

If it is deemed necessary to make an exception and enable the feature, admins should limit the ability to publish to the web to only specific security groups, instead of allowing the entire agency to publish data to the web."

This would be updated to the following:
"1. Publish to Web

This section helps reduce security risks related to Power Bi's capability to publish to the web. "

This was a note by Ted since a lot of the text will be covered by the rationale and is inconsistent across baselines.

ahuynhMITRE:
> 

good point, will add as a parking lot item for tomorrow. Probably lost in the many conversations / reviews over the past months and open to whatever works best.

ahuynhMITRE (approved):
all good after the discussion at the stand up!

nanda-katikaneni:
@amart241 @buidav - seems there are conflicts for this branch to merge into emerald, can you please update/rebase and resolve conflicts.

PR #441, opened 2023-07-17
Update MS.AAD.3.2v1 & MS.AAD.3.6v1 to Check Authentication Strength for Phishing Resistance MFA

### LINKED ISSUE #409: Update AAD MS.AAD.3.2v1 to be dependent on MS.AAD.3.1v1
# 💡 Summary #

Update the Rego code for MS.AAD.3.2v1 If Phishing-resistant MFA has not been enforced yet, then an alternative MFA method SHALL be enforced for all users to be dependent on MS.AAD.3.1v1 Phishing-resistant MFA SHALL be enforced for all users.

- If policy MS.AAD.3.1v1 Passes, then this policy MS.AAD.3.2v1 will automatically Pass. See the way the OneDrive expiration date for Anyone links is dependent on a different setting for an example.
- If policy MS.AAD.3.1v1 Fails, then the existing Rego code for MS.AAD.3.2v1 should be applied to evaluate the policy (i.e. checking for a conditional access policy that requires MFA but does not dictate the specific type of MFA). [See the instructions section](https://github.com/cisagov/ScubaGear/blob/emerald/baselines/aad.md#msaad32v1-instructions) as a reference for the conditional access policy template but again this code already exists in the Rego.

 dependsOn #433 

## Acceptance criteria ##

How do we know when this work is done?

- [ ] Test the code for all possible value combinations of MS.AAD.3.1v1 and MS.AAD.3.2v1 to ensure that the report produces an accurate Pass or Fail
- [ ] Test in all tenants

### LINKED ISSUE #413: Update ScubaGear Rego check for MS.AAD.3.6v1 MFA required for Highly Privileged Roles to check for phishing-resistant MFA
# 💡 Summary #

Update the existing Rego code for MS.AAD.3.6v1 MFA for highly privileged roles to check for a conditional access policy that requires a phishing-resistant MFA method.

- There is already code that checks for MFA. For the most part that should stay the same, but the logic needs to be updated to check for Authentication Strength > Phishing-resistant MFA. See the [conditional access policy template in the baseline instructions](https://github.com/cisagov/ScubaGear/blob/emerald/baselines/aad.md#msaad36v1-instructions) as a reference.

dependsOn #433 

## Acceptance criteria ##

How do we know when this work is done?

- [ ] Test to ensure that the code produces a Pass or Fail as expected for various Compliant/NonCompliant states of conditional access policies.
- [ ] Test against all tenants.

crutchfield:
#  #

## 🗣 Description ##

Check Authentication Strength for Phishing Resistance MFA.
Make both MS.AAD.3.2v1 and MS.AAD.3.6v1 depend/use MS.AAD.3.1.1v1 Authentication Strength checks.  

closes #409 
closes #413

Phisting resistant MFA is required in new secure baseline for MS.AAD.3.2v1 and MS.AAD.3.6v1

## 🧪 Testing ##

Added additional unit tests for MS.AAD.3.2v1 and MS.AAD.3.6v1.  
Ran InvokeScuba against all tenants. 

image

image

- [X] This PR has an informative and human-readable title.
- [X] Changes are limited to a single goal - *eschew scope creep!*
- [X] *All* future TODOs are captured in issues, which are referenced
      in code comments.
- [X] All relevant type-of-change labels have been added.
- [X] All relevant repo and/or project documentation has been updated
- [X] Tests have been added and/or modified to cover the changes in this PR.
- [X] All new and existing tests pass (except with open issues).

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Create a release.

### DIFF
--- a/Rego/AADConfig.rego
+++ b/Rego/AADConfig.rego
@@ -371,7 +371,7 @@ tests[{
     "RequirementMet" : false
 }] {
     PolicyId := "MS.AAD.3.3v1"
-    count(MS_AAD_3_1v1_CAP) = 0
+    count(MS_AAD_3_1v1_CAP) == 0
 }
 #--
 
--- a/Rego/AADConfig.rego
+++ b/Rego/AADConfig.rego
@@ -93,8 +93,8 @@ ReportDetailsBooleanLicenseWarning(_) = Description if {
 # User/Group Exclusion support functions #
 ##########################################
 
-default UserExclusionsFullyExempt(_, _) = false
-UserExclusionsFullyExempt(Policy, PolicyID) = true if {
+default UserExclusionsFullyExempt(_, _) := false
+UserExclusionsFullyExempt(Policy, PolicyID) := true if {
     # Returns true when all user exclusions present in the conditional
     # access policy are exempted in matching config variable for the
     # baseline policy item.  Undefined if no exclusions AND no exemptions.
@@ -104,14 +104,14 @@ UserExclusionsFullyExempt(Policy, PolicyID) = true if {
     count(ExcludedUsers - AllowedExcludedUsers) == 0
 }
 
-UserExclusionsFullyExempt(Policy, PolicyID) = true if {
+UserExclusionsFullyExempt(Policy, PolicyID) := true if {
     # Returns true when user inputs are not defined or user exclusion lists are empty
     count({ x | x := Policy.Conditions.Users.ExcludeUsers[_] }) == 0
     count({ y | y := input.scuba_config.Aad[PolicyID].CapExclusions.Users }) == 0
 }
 
-default GroupExclusionsFullyExempt(_, _) = false
-GroupExclusionsFullyExempt(Policy, PolicyID) = true if {
+default GroupExclusionsFullyExempt(_, _) := false
+GroupExclusionsFullyExempt(Policy, PolicyID) := true if {
     # Returns true when all group exclusions present in the conditional
     # access policy are exempted in matching config variable for the
     # baseline policy item.  Undefined if no exclusions AND no exemptions.
@@ -121,7 +121,7 @@ GroupExclusionsFullyExempt(Policy, PolicyID) = true if {
     count(ExcludedGroups - AllowedExcludedGroups) == 0
 }
 
-GroupExclusionsFullyExempt(Policy, PolicyID) = true if {
+GroupExclusionsFullyExempt(Policy, PolicyID) := true if {
     # Returns true when user inputs are not defined or group exclusion lists are empty
     count({ x | x := Policy.Conditions.Users.ExcludeGroups[_] }) == 0
     count({ y | y := input.scuba_config.Aad[PolicyID].CapExclusions.Groups }) == 0
@@ -134,8 +134,8 @@ GroupExclusionsFullyExempt(Policy, PolicyID) = true if {
 #
 # MS.AAD.1.1v1
 #--
-default LegacyAuthenticationConditionsMatch(_) = false
-LegacyAuthenticationConditionsMatch(Policy) = true if {
+default LegacyAuthenticationConditionsMatch(_) := false
+LegacyAuthenticationConditionsMatch(Policy) := true if {
     "All" in Policy.Conditions.Users.IncludeUsers
     "All" in Policy.Conditions.Applications.IncludeApplications
     "other" in Policy.Conditions.ClientAppTypes
@@ -176,8 +176,8 @@ tests[{
 #
 # MS.AAD.2.1v1
 #--
-default BlockHighRiskConditionsMatch(_) = false
-BlockHighRiskConditionsMatch(Policy) = true if {
+default BlockHighRiskConditionsMatch(_) := false
+BlockHighRiskConditionsMatch(Policy) := true if {
     "All" in Policy.Conditions.Users.IncludeUsers
     "All" in Policy.Conditions.Applications.IncludeApplications
     "high" in Policy.Conditions.UserRiskLevels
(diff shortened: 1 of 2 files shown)


ssatyapal123 (changes requested):
Ran against E5, y2zj1 and got the expected result:
image

Then I used Ted's non-invasive testing to edit the policy for "Live - MFA SHALL be required for all users" in the test ProviderSettingsExport.json, by changing, "State" to "Off" and "Users included" to "None", but the test still showed as a pass:
image

tkol2022:
> @tkol2022 Reviewing this item, I'm wondering if there may be two possible interpretations of the policies that affect the assessment. One assessment would only check that the current policy state is such that MFA is being enforced consistent with either 3.1 or 3.2 and 3.6. But the rationales for 3.2 and 3.6 indicate that they are meant as backup policies in case 3.1 were to be disabled (perhaps are part of changes or an operational misconfiguration) or nonexistent (due to an agency not yet being ready to widely support phishing resistant MFA broadly for all users). So, my question is... which is it? Because the assessment would work out differently depending on the which interpretation you use.
> 
> Scenario: Only checking if all users are covered with MFA Conditions: Phishing-resistant MFA enabled for all users, no backup policies for 3.2 or 3.6 present. Expected Results: 3.1: Pass 3.2: Pass 3.6: Pass
> 
> Scenario: Checking if backup policies are in place in case of future misconfigurations or purposeful adversary disablement Conditions: Phishing-resistant MFA enabled for all users, no backup policies for 3.2 or 3.6 present. Expected Results: 3.1: Pass 3.2: Fail 3.6: Fail
> 
> I think, in order to determine if this code is in keeping with the policies, we have to clarify the intent a d which scenario then applies.

I'm open to ideas on whatever you all think makes the most sense. Below are my thoughts. I'm adding 3.3 to the discussion because that policy also depends on 3.1.

- Any policy in section 3 that has a name which starts with "If phishing-resistant MFA has not been enforced yet", depends on policy 3.1. See the next set of bullets for details.
- Policy 3.2 is ONLY necessary in the event that 3.1 is not implemented. If the agency has implemented phishing-resistant MFA and applied it to all users, then 3.2 is redundant. The real reason we kept this policy is because we know that in practice, it is going to take time for agencies to become fully phishing-resistant across all of their users. If 3.1 passes, then we can automatically pass 3.2.
- Policy 3.3 is also only necessary in the event that 3.1 is not implemented and the tenant has Microsoft Authenticator enabled. Just like for 3.2, we created this policy as a stop-gap measure to help further secure tenants that will take time to transition to full phishing-resistant MFA. Also, non governmental organizations such as the commercial sector or education sector may not be able to deploy PIV cards or FIDO tokens, so the next best thing for them is MS Authenticator configured in a way that helps reduce modern phishing attack techniques. If 3.1 passes, then we can automatically pass 3.3.
- Policy 3.6 is a little bit different. This one is truly meant to be a redundant policy and is NOT dependent on 3.1. From an assessment standpoint, we expect both 3.1 and 3.6 to be implemented and 3.6 should be assessed independently of 3.1. The thinking behind 3.6 is: for tenants that cannot fully deploy phishing-resistant MFA to ALL of their users, they should at least make phishing-resistant MFA required for administrators. Also, since 3.6 is a backup policy, in the event that 3.1 was ever misconfigured, at least phishing-resistant MFA would still be enforced for privileged admin users.
- Policies 3.2 and 3.3 are not really "backup policies" in the same way that 3.6 is. I would say 3.2 and 3.3 are more like stop-gap policies in the event that 3.1 is not deployed yet. Policy 3.6 is truly a backup (redundant) policy. Because of this I'm thinking we might want to update the rationales for 3.2 and 3.3 if you think it makes sense?

Let me know if this makes sense and we can also discuss verbally if necessary.

ssatyapal123:
@tkol2022 @schrolla Ted asked me to review the wording and apologize if I'm oversimplifying but it seemed to me the best way forward was simply to replace the word "backup" in the Rationale for MS.AAD.3.2v1 and MS.AAD.3.3v1.

Ted recommended the word "stopgap". Another option is "substitute", however I think stopgap is better because it implies this option is not as good as the phishing resistant option.

Another point I noticed as I was reviewing the tech editor comments is that the subheadings closely resemble but don't match the control numbers, e.g.:

3.**1**.6 MS.AAD.3.6v1

I got a bit confused by this and wondered if it would be better to change this level of subheadings to use a different scheme, such as

a) or i), or "-"

crutchfield:
@schrolla  I left 3.2 as is.  For 3.3 - if 3.1 passes it passes, otherwise we report "Not checked".  For 3.6, treated as backup policy so I removed condition of 3.1 passing automatically passing 3.6.   Unit test updated to check these conditions.

crutchfield:
> One minor correction requested. As an aside, why the change of all assignments to use the unification operator? I don't think, in this case, it results in a functional difference but it is counter to the Rego policy reference which states that, "Best practice is to use assignment `:=` and comparison `==` wherever possible."
> 
> So, if there isn't a specific reason you are using unification('='), I would recommend against it in case of unintended side effects since it bundles both assignment and comparison into one and it appears in nearly all cases, you are looking to perform assignment here.
> 
> https://www.openpolicyagent.org/docs/latest/policy-language/#best-practices-for-equality

I had my VS Code plug in pointing to previous version of OPA Agent (0.38.1) and it reports "c:/Users/richa/source/repos/ScubaGear/Rego/DefenderConfig.rego:30: rego_parse_error: functions must use = operator (not := operator)"

With this updated (pint plug-in to actual one downloaded for Scuba) to the most recent version this issue goes away.  I am changing back to :=.  But note, the changes were to functions and not assignments or camparisions.  But I cannot find the "best practice for using "=" in functions in any current documentatipn from OPA.

nanda-katikaneni:
Tested on G5 - merging into Emerald.

PR #455, opened 2023-07-25
Formal internal review of PowerBI baseline

schrolla:
#  #

## 🗣 Description ##

This PR is designed to facilitate a formal review of the revised PowerBI baseline from CISA SCuBA principals and the Scuba development team. Following this review the baseline should be in a state for sending to the CISA agency formal review process.

      in code comments.

## ✅ Pre-merge checklist ##

- [ ] Resolve conflicts with base branch via rebase.

### DIFF
--- a/baselines/powerbi.md
+++ b/baselines/powerbi.md
@@ -1,6 +1,6 @@
 # CISA M365 Security Configuration Baseline for Power BI
 
-Power BI is a Software as a Service (SaaS) offering from Microsoft thatfacilitates self-service business intelligence dashboards, reports, datasets, and visualizations. Power BI can connect to multiple, different data sources, combine and shape data from those connections, then create reports and dashboards to share with others. This Secure Configuration Baseline (SCB) provides specific policies to strengthen Power BI security.
+Power BI is a Software as a Service (SaaS) offering from Microsoft that facilitates self-service business intelligence dashboards, reports, datasets, and visualizations. Power BI can connect to multiple, different data sources, combine and shape data from those connections, then create reports and dashboards to share with others. This Secure Configuration Baseline (SCB) provides specific policies to strengthen Power BI security.
 
 The Secure Cloud Business Applications (SCuBA) project run by the Cybersecurity and Infrastructure Security Agency (CISA) provides guidance and capabilities to secure federal civilian executive branch (FCEB) agencies’ cloud business application environments and protect federal information that is created, accessed, shared, and stored in those environments. 
 
@@ -185,7 +185,7 @@ on roles.
 #### MS.POWERBI.3.1v1
 The Invite external users to your organization feature SHOULD be disabled unless agency mission requires the capability.
 
-- _Rationale:_ Allowing guest users in general is a bad idea and can lead to unwanted access. Disabling this feature in PowerBI is just another way to avoid having a risky guest user in the organization. If external invites are required then, guest users should be invited by authorized users via the Azure Active Directory guest invite features. This helps ensure that authorized administrative personnel with the right permissions can centrally manage who has access to the tenant.
+- _Rationale:_ Disabling this feature in PowerBI is just another way to avoid having a risky guest user in the organization. If external invites are required then, guest users should be invited by authorized users via the Azure Active Directory guest invite features. This helps ensure that authorized administrative personnel with the right permissions can centrally manage who has access to the tenant.
 - _Last modified:_ June 2023
 
 ### Resources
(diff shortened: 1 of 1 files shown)


tkol2022:
### MS.POWERBI.1.1v1 instructions
If you follow step 1 to open the Power BI admin portal, then steps 2 and 3 are unnecessary. I would remove them unless there is a reason to have them there. I was able to configure the setting as described in steps 4-6 without them.

tkol2022:
### Length and content of introduction paragraphs not consistent with other baselines
The three intro paragraphs at the top of the baseline, in particular paragraphs 2 and 3, read more like an "Intro to Power BI" primer or Microsoft general documentation page. Although that information might be useful, the other baselines "cut to the chase" and do not contain this type of language. I suggest trimming that down and only keeping content that is absolutely necessary to help understand any of our policies.

tkol2022:
### 2. Power BI Guest Access intro section
The section 2 introduction has some issues that I'd like to see addressed:

1. "A best practice is to disallow guest user access. Disallowing guest access also aligns with zero trust principles." It is not clear to me how disallowing guest access in Power BI aligns with zero trust principles. In AAD we allow guest user access. Is there something specific in Power BI that makes guest users high risk? Guest users also need to follow the same authentication and access control requirements as normal tenant users in AAD so from a security standpoint the only difference is that guests belong to different organizations.
2. "Enabling this setting allows AAD Business-to-Business (AAD B2B) guest users to access Power BI...Using the specific security groups option allows administrators to control which guest users can access Power BI. " The second paragraph described a specific setting but section intros should be generally about the settings in that section. Language about specific policies should be include inside the policies themselves, not the intro.
3. The intro paragraphs are long and seem to have extraneous content. Sentences like "If this setting is disabled, guest users receive an error when trying to access Power BI. Disabling this setting for the entire agency also prevents users from inviting guests to the agency" seem extraneous to me and I would consider removing them. I also cannot see why we need to describe internal users, external users, B2B guest users since these terms don't seem directly related to the policy - I'm not sure that they add anything there.
4. The Note in the intro is really about policy 2.1 and should be moved to that section. It is not intro material.
5. "If the Power BI tenant’s guest users will need to own and share Power BI content with others and manage workspaces as Power BI workspace Admins, Microsoft recommends changing the Guest user permissions are limited setting in AAD to allow these users to use people pickers within the Power BI UX. Since Power BI integrates natively with AAD, the AAD Baseline should be consulted for additional guidance on managing guest users." First, policy 2.1 instructs to Disable Guest access so if guest access is disabled, then is this guidance only applicable when an an agency is implementing an exception to the policy?  Second, if the user does need to keep Guest access Enabled, then it is unclear what we are recommending the user to do? "Microsoft recommends changing the Guest user permissions...to allow..." Are we telling the user to change the value of any of the configurations in Section 8 of the AAD baseline since that deals with user access?

tkol2022:
### MS.POWERBI.2.1v1 instructions
The name of the setting step 6 looks like it changed to **Allow Azure Active Directory guest users to edit and manage content in the organization**.

tkol2022:
### MS.POWERBI.3.1v1

- I have the same question about this policy as I did about 2.1. See comment 1 for section 2. Why are we not allowing Guest Access?

- The policy has the text "Invite external users to your organization" in bold. Remove the bold to be consistent with our other policies.

tkol2022:
### MS.POWERBI.4.1v1 instructions

- Based on my recommended policy change to 4.1, put a note at the top of the instructions saying that this configuration is only necessary if the Power BI setting **Allow service principals to use Power BI APIs** is enabled. Otherwise the user doesn't need to perform any configurations because service principals would be off.
- Step 7 is a second configuration "Allow service principals to create and use profiles" and is not explained anywhere up above in the policy section. Need to explain what this setting does and why it is necessary (rationale).

tkol2022:
### MS.POWERBI.6.1v1

- I'm not sure that completely turning off Python and R seems like a reasonable recommendation. I would think that for a tool like Power BI an agency would want to be able to integrate tools like Python Pandas which is a common data analysis tool. Is there a secure way to integrate data analysis tools?

tkol2022:
### MS.POWERBI.7.1v1

- Recommend changing the policy to "Users SHALL be allowed to apply sensitivity labels for Power BI content" to better align with what the configuration actually is.

nanda-katikaneni (commented):
Completed my review - overall in good shape with a few minor changes requested (please look for comments near where the changes are suggested).

tkol2022 (approved):
I'm approving based on all the updates made that align with my comments and the meeting we had on 9/1/2023. Nice work. See my email for one last revision that might be needed.

ahuynhMITRE:
@amart241 moved a few things around based on feedback from Grant, this includes:

- moving guest definitions to key terminology 
- moved best practices for service accounts to implementation considerations appendix

ahuynhMITRE:
@nanda-katikaneni good to merge

PR #464, opened 2023-07-28
Formal internal review of Exchange Online baseline

schrolla:
#  #
## 🗣 Description ##

This PR is designed to facilitate a formal review of the revised Exchange Onlinebaseline from CISA SCuBA principals and the Scuba development team. Following this review the baseline should be in a state for sending to the CISA agency formal review process.

      in code comments.
- [x] All new and existing markdown tests pass.

## ✅ Pre-merge checklist ##

- [ ] Feature branch has been rebased against base branch with conflicts resolved

## ✅ Post-merge checklist ##

- [ ] Ensure Exchange Online related markdown tests pass on merged branch

### DIFF
--- a/baselines/exo.md
+++ b/baselines/exo.md
@@ -13,6 +13,12 @@ this baseline. When noted, alternative products may be used in lieu of
 Defender, on the condition that they fulfill these required baseline
 settings.
 
+The Secure Cloud Business Applications (SCuBA) project run by the Cybersecurity and Infrastructure Security Agency (CISA) provides guidance and capabilities to secure federal civilian executive branch (FCEB) agencies’ cloud business application environments and protect federal information that is created, accessed, shared, and stored in those environments.
+
+The CISA SCuBA SCBs for M365 help secure federal information assets stored within M365 cloud business application environments through consistent, effective, and manageable security configurations. CISA created baselines tailored to the federal government’s threats and risk tolerance with the knowledge that every organization has different threat models and risk tolerance. Non-governmental organizations may also find value in applying these baselines to reduce risks.
+
+The information in this document is being provided “as is” for INFORMATIONAL PURPOSES ONLY. CISA does not endorse any commercial product or service, including any subjects of analysis. Any reference to specific commercial entities or commercial products, processes, or services by service mark, trademark, manufacturer, or otherwise, does not constitute or imply endorsement, recommendation, or favoritism by CISA.
+
 ## License Compliance and Copyright
 
 Portions of this document are adapted from documents in Microsoft’s
@@ -250,9 +256,12 @@ The DMARC point of contact for aggregate reports SHALL include <reports@dmarc.cy
 - _Rationale:_ Email spoofing attempts are not inherently visible to domain
 owners. DMARC provides a mechanism to receive reports of spoofing attempts.
 Including <reports@dmarc.cyber.dhs.gov> as a point of contact for these reports
-gives CISA insight into spoofing attempts and is required by Binding Operational Directive 18-01, "Enhance Email and Web Security” for
-federal, executive branch, departments and agencies.
+gives CISA insight into spoofing attempts and is required by Binding Operational
+Directive 18-01, "Enhance Email and Web Security” for federal, executive
+branch, departments and agencies.
 - _Last modified:_ June 2023
+- _Note:_ Only federal, executive branch, departments and agencies should
+          include this email address in their DMARC record.
 
 #### MS.EXO.4.4v1
 An agency point of contact SHOULD be included for aggregate and failure reports.
@@ -306,10 +315,12 @@ PowerShell tool `Resolve-DnsName`. For example:
 Resolve-DnsName _dmarc.example.com txt
 ```
 
-If DMARC is configured, a response resembling `v=DMARC1; p=reject; pct=100; rua=mailto:reports@dmarc.cyber.dhs.gov, mailto:reports@example.com`
-will be returned, though by necessity, the contents of the record will vary by agency.
-In this example, the policy indicates that all emails that fail the SPF/DKIM checks are to
-be rejected and reported to reports@dmarc.cyber.dhs.gov and reports@example.com.
+If DMARC is configured, a response resembling `v=DMARC1; p=reject; pct=100; rua=mailto:reports@dmarc.cyber.dhs.gov, mailto:reports@example.com; ruf=mailto:reports@example.com`
+will be returned, though by necessity, the contents of the record will vary
+by agency. In this example, the policy indicates that all emails that fail the
+SPF/DKIM checks are to be rejected and aggregate reports sent to
+reports@dmarc.cyber.dhs.gov and reports@example.com. Failure reports will be
+sent to reports@example.com.
 
 #### MS.EXO.4.2v1 instructions:
 See [MS.EXO.4.1v1 instructions](#msexo41v1-instructions) for an overview of how to publish and check a DMARC record. Ensure that the record published includes `p=reject`.
@@ -321,7 +332,8 @@ as one of the emails for the `rua` field.
 #### MS.EXO.4.4v1 instructions:
 See [MS.EXO.4.1v1 instructions](#msexo41v1-instructions) for an overview of how to publish and check a DMARC record. Ensure that the record published includes a point of contact
 specific to your agency in addition to <reports@dmarc.cyber.dhs.gov>
-as one of the emails for the `rua` field.
+as one of the emails for the `rua` field and only one or more agency-defined points
+of contact for the `ruf` field.
 
 ## 5. Simple Mail Transfer Protocol Authentication (SMTP AUTH)
 
--- a/baselines/exo.md
+++ b/baselines/exo.md
@@ -7,7 +7,7 @@ Many admin controls for Exchange Online are found in the **Exchange admin center
 However, several of the
 security features for Exchange Online are shared between Microsoft
 products and are configured in either the **Microsoft 365 Defender portal**
-or **Microsoft 365 Purview compliance portal** admin centers. Generally
+or **Microsoft 365 Purview compliance portal**. Generally
 speaking, the use of Microsoft Defender is not strictly required for
 this baseline. When noted, alternative products may be used in lieu of
 Defender, on the condition that they fulfill these required baseline
@@ -1019,7 +1019,7 @@ OMB M-21-13 also requires Advanced Audit be configured in M365. Advanced Audit,
 now Microsoft Purview Audit (Premium), adds additional event types to the
 Unified Audit Log.
 
-Audit logging is managed from the Microsoft Purview compliance center. For
+Audit logging is managed from the Microsoft Purview compliance portal. For
 implementation guidance for configuring audit logging, see the following
 section of the CISA M365 Security Configuration Baseline for Defender for
 Office 365.
--- a/baselines/exo.md
+++ b/baselines/exo.md
@@ -555,6 +555,14 @@ The DLP solution SHALL protect PII and sensitive information, as defined by the 
 
 - None
 
+### Implementation
+
+#### MS.EXO.8.1v1 instructions
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [DLP](./defender.md#implementation-3) for additional guidance.
+
+#### MS.EXO.8.2v1 instructions
+Any product that meets the requirements outlined in this baseline policy may be used. If the agency is using Microsoft Defender, see the following implementation steps for [protecting PII](./defender.md#msdefender41v1-instructions) for additional guidance.
+
 ## 9. Attachment File Type
 
 For some types of files (e.g., executable files), the dangers of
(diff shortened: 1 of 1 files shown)


baselines/exo.md
```diff
@@ -994,18 +1000,29 @@ The alerts SHOULD be sent to a monitored address or incorporated into a SIEM.
...
 ## 17. Microsoft Purview Audit
 
-Unified audit logging generates logs of user activity in M365 services.
-These logs are essential for conducting incident response and threat detection activity.
+User activity from M365 services are captured in the organization's unified
+audit log.  These logs are essential for conducting incident response and
+threat detection activity.
 
-By default, Microsoft retains the audit logs for only 90 days. Activity by users with E5 licenses is logged for one year.
+By default, Microsoft retains the audit logs for only 90 days. Activity by
```
gdasher, line 1007: Please update this per the recent announcement to note that this will be extended to 180 days and will expand to include purview premium events as well.

https://www.microsoft.com/en-us/security/blog/2023/07/19/expanding-cloud-logging-to-give-customers-deeper-security-visibility/

baselines/exo.md
gdasher, line 1: MS.EXO.5.2v1: typically, we have put the "MAY" into the policy with the "SHALL NOT".  Shoudl this be added to MS.EXO.5.1v1 instead for consistently with the approach in other policies from other baselines?

gdasher: Sorry I missed the standup.  Yes, makes sense.    I had not reviewed Defender yet hence I assumed the other format was standard...

tkol2022:
### 17 Microsoft Purview Audit
The following are suggestions / questions related to the audit policy section.

- Suggest changing the name of the section to something more generic like Audit Logs or something like that instead of naming it after Microsoft's feature name.
- "Activity by users with E5 licenses is logged for one year." Did you mean to say with E5 licenses is **retained** for one year?
- MS.EXO.17.3v1 audit retention policy - In theory I understand why we have this policy, but I'm struggling with it because it doesn't seem that there is a single, standard way to "configure" it, like our other hardening policies. Technically an agency could offload their logs to a SIEM and then the SIEM would be responsible for retaining the logs per the OMB period - is that correct? If that is true, then an agency wouldn't need to follow our implementation instructions in Defender to configure and Audit Retention Policy within M365 because they would be handling that in a separate system. I think this policy might fit into the category of "the user can implement the requirement with an alternative system", which in this case would be a separate log retention system. Not sure how my discussion here changes our policy but I wanted to discuss it nonetheless.
- "Audit logs SHALL be maintained for at least the minimum duration dictated by OMB M-21-31 (Appendix C)]" - Appendix C contains a lot of different types of logs and it is not obvious which ones correspond to the log types described within baseline section 17. Consider pointing the user to the specific log types which are relevant. I know that we mention the "Unified Audit Log" when discussing OMB in policies 17.1 and 17.2 but I think 17.3 should also make it clear what to look for in the OMB memo retention table.
- Question about audit log retention in general. Section 17 seems to be scoped to the Unified Audit Log. What about all the other log types that we reference in AAD section 4 Centralized Log Collection? For completeness, do the baselines need an audit retention policy for those AAD logs as well?
- The implementation section is not included in this baseline but it is included in Teams for policies that point to Defender. In Teams, the implementation section it is simply a sentence instructing the user to review the relevant Defender section with a hyperlink (see screenshot below). I recommend to do the same thing with EXO. I personally think it is better to have an Implementation section even if it is just to point the user to the relevant Defender section.

image

tkol2022:
### MS.EXO.1.1v1 instructions
In the instructions for the automatic email forwarding policy, it only references the "Default" domain policy. ScubaGear on the other hand, checks the Default and any other policies to ensure that they all have Allow Automatic Forwarding turned Off. I recommend augmenting the instructions to include the possibility that the user may have multiple policies in addition to just the Default. Maybe we just need to say, "repeat these instructions for each Remote Domain policy" or something like that.

image
image

tkol2022:
### MS.EXO.2.1v1
Recommend updating the policy name to be more descriptive of what the actual hardening configuration is - maintaining a list in an of itself isn't really what the intent is here - needs more details. How about something like: "A list of approved IP addresses for sending mail SHALL be maintained in agency's DNS SPF records"?

I also recommend updating the instructions for this policy to make it more obvious what the user is supposed to configure. For example, it instructs the user to "Identify any approved senders specific to your agency." Then what is the user supposed to do with the list of approved senders? 
Also the word "senders" in the 2.1 instructions is not clear. Is that supposed to be "IP addresses" as referenced in the policy?

When I read the MS.EXO.2.2v1 instructions I get the impression that it covers both policies. In fact policy 2.1 is referenced in those instructions. Maybe it would be better to simply have a single instructions section where both policies 2.1 and 2.2 are configured together. Perhaps our standard approach of making each instruction set atomic doesn't apply in this case?

tkol2022:
### MS.EXO.3.1v1 DKIM SHOULD be enabled for any custom domain

Why is this a SHOULD policy whereas the SPF and DMARC policies are SHALL? The respective DHS BOD includes DKIM as well.

tkol2022:
### MS.EXO.4.3v1
This is a policy that is truly government specific because it instructs to set the value to reports@dmarc.cyber.dhs.gov which clearly won't apply to non FCEB organizations. I'm wondering if we should create a special note field to call this out, or simply include it in the notes.

### MS.EXO.4.4v1
This is a nitpick. This policy "An agency point of contact SHOULD be included for aggregate **and failure reports**" is slightly different from 4.3 because 4.3 does not include the words **and failure reports**. If there is a reason that they differ, then do nothing. If they are supposed to be the same, then go ahead and make the language identical for consistency.

tkol2022:
### MS.EXO.5.1v1 SMTP AUTH SHALL be disabled in Exchange Online.
Not sure if the words "Exchange Online" are needed in the policy name. It might be better to remove them, unless we have standardized the usage of "in Exchange Online" with all the other policies and there is some reason for it there.

### MS.EXO.5.2v1 SMTP AUTH MAY be enabled on a per-mailbox basis as needed.
To re-iterate my analysis of these types of policies, here you go. I believe that this policy is really more of an exception of policy 5.1 and not really its own hardening policy. Also the only reason I would consider keeping any references to SMTP auth is if we know of a way to configure it "securely" for specific exception mailboxes. If not, I think we should remove all references to it including any respective instructions because I don't think it contributes to the security of the tenant, at least not in any obvious way.

ahuynhMITRE (commented):
no additional comments, all of mine were either covered by Ted or Grant

tkol2022:
### MS.EXO.6.1v1 and MS.EXO.6.2v1
_Please read all of the comments before making any changes because there are some interdependencies._

- I'm wondering if these policies should be rephrased as "Contact folders SHALL only be shared with approved domains" instead of "Contact folders SHALL NOT be shared with all domains". When I go to configure the policy in the admin portal, if I select "Share your Contacts folder", it forces me to enter a list of approved domains (see screenshot below). So technically I'm not sure there is a way that I can just turn Off contact folder sharing, however look at my third comment bullet below as well because I did potentially find another way I could configure the admin portal and still be compliant with the policy.
- For 6.1 instructions step 5, I'm wondering if we augment the instructions to say "For all sharing rules under all existing policies where "Share your contacts folder" is selected, ensure Sharing with all domains is not selected". Same tweak for 6.2 instructions but it would say where "Share your calendar folder" is selected. See the respective checkboxes in the screenshot below.

image

- I was able to also configure the settings as per the screenshot below by selecting **Sharing with all domains** but then deselecting both **Share your contacts folder** and **Share your calendar folder**. I believe this might achieve the purpose of the baseline policy but I haven't hands-on tested it. If this does in fact achieve the purpose of the policy, then disregard my first comment bullet. Maybe we need to augment the instructions to include this configuration?

image

tkol2022:
### MS.EXO.7.1v1 instructions

- Step 3 assumes that the user doesn't already have an existing rule.
- After step 12 I recommend adding what you want the **Match sender address in message** value to be set to. The default for me was the value **Header**.

image

- After I created the rule as described in the instructions, it shows as Disabled in the admin portal. Could there be steps missing in the instructions? For my test rule, I had to **set a specific activation time first**, and then go back in and set the **Enable or disable rule** field to Enabled.

image

image

image

tkol2022:
### 8. Data Loss Prevention Solutions
I'm wondering if we can combine policies 8.1 and 8.2 into a single policy. I don't see 8.1 as being specific hardening guidance - rather it states a high level requirement, that a DLP system must be used - this deviates from other policies which are specific. Maybe we can have a single policy that reads something like:

`A DLP solution SHALL be used that protects PII and sensitive information, as defined by the agency. At a minimum, the sharing of credit card numbers, Taxpayer Identification Numbers (TIN), and Social Security Numbers (SSN) via email SHALL be restricted.
`

Also, the second sentence in policy 8.1 is not specific: "The selected DLP solution SHOULD offer services comparable to the native DLP solution offered by Microsoft." What specific services is this policy setting as the requirement? Is the protection of PII credit cards, TINs and SSNs via email, the only requirements? Are there more?

tkol2022:
### MS.EXO.9.1v1
Here's a suggested update to the policy.

`There SHALL be an email filtering policy that filters attachments for risky file types. The selected filtering solution SHOULD offer services comparable to Microsoft Defender's Common Attachment Filter.`

Also:

- The policy mixes SHALL and SHOULD which makes the criticality ambiguous, at least to me. Is the second sentence really a SHOULD?
- The second sentence for 9.1 seems like it would apply to 9.2 and 9.3 as well. Maybe it is better to move that sentence to the intro paragraph if it applies broadly to all these policies since they all see to be about attachment file types and filtering?

tkol2022:
### MS.EXO.10.3v1
Suggestion wording update to the policy for clarity below:

`Email scanning SHOULD be capable of retroactively reviewing emails after they have already been delivered.`

Also, I would consider changing the criticality to a SHALL since this seems like a really important feature to me.

tkol2022:
### License Requirements
I noticed that some policies do not contain the license requirements section (e.g. section 11 Phishing Protections). Is this because those policies point to Defender and the Defender baseline is responsible for stating those requirements?

tkol2022:
### MS.EXO.11.1v1 instructions

The hyperlink to Defender points to the wrong section. I would think it should point to the implementation instructions for section 2 Impersonation Protection.

tkol2022:
### Section 11 Phishing Protections licensing confusion

The following sentences were confusing: "Microsoft Defender has capabilities for all of these phishing protections. With the exception for impersonation protection, these features are available with Exchange Online Protection (EOP), which is included in all Microsoft 365 subscriptions that contain Exchange Online mailboxes."

- What exactly are we trying to point out here? That if you want Impersonation Detection you need to have Defender? Consider rewording as:

`These feature described in this policy section are available with Exchange Online Protection (EOP), which is included in all Microsoft 365 subscriptions that contain Exchange Online mailboxes. The only exception is impersonation protection, which is included if you subscribe to Microsoft Defender for Office 365.
`

- Also, I would update all references to "Microsoft Defender" to be specific to the Defender flavor relevant to this baseline.

- Not sure, but maybe this content belongs in the License Requirements section?

tkol2022:
### 12. IP Allow Lists Implementation

With the exception of the DKIM policy, this is the only section in the entire baseline that instructs the user to go to the Microsoft 365 Defender Portal to implement the policies. Should this implementation be moved to Defender, like we did for other policies?

image

tkol2022:
### MS.EXO.12.1v1 instructions and MS.EXO.12.2v1 instructions

Step 8 for 12.1 and step 7 for 12.2 seem to be duplicates of each other.

image

tkol2022:
### 13. Mailbox Auditing

- "Some actions performed by administrators, delegates, and owners are logged automatically. " Not sure about the purpose of this sentence. Logged automatically by what? What does that have to do with the Mailbox Auditing feature?

tkol2022:
### MS.EXO.14.3v1

Policy states:

`Allowed domains SHALL NOT be added.`

Shall not be added to what? Be specific.

tkol2022:
15. Link Protection

- Missing period between sentence. "Microsoft Defender includes link scanning capabilities Use of Microsoft Defender is not strictly required for this purpose;"
- MS.EXO.15.2v1 "Direct download links SHOULD be scanned for malware." Since section 10 is dedicated to malware, what is the reason that 15.2 is in this section? I'm fine with it - I just want to know if there is a difference between the malware policies in section 10 versus other sections.

tkol2022:
### 16. Alerts

- This section seems to be heavily influenced by Defender. It almost seems like it not necessarily generic (i.e. could be implemented by another product). I'm not saying that it cannot be implemented by something else - just that it appears as such based on the language used. For example the first paragraph starts with "Microsoft Defender includes several prebuilt alert policies, many of which pertain to Exchange Online." Nothing else in the intro states that a competing product can be used like in some of the other EXO policies. Policy 16.1 contains a specific list of alerts, including one named "Suspicious Connector Activity" which seems like it might be Defender-specific.
- The instructions section has some invalid links to the Defender document.

tkol2022:
I have completed my review of this baseline. Thanks for taking my questions and suggestions. I plan to go back over any specific policies that you want me to re-review (if necessary) and also help adjudicate any open questions if necessary by the end of this week. Nice work and let me know if there are any high priority discussion topics to close out on this baseline.

tkol2022 (approved):

gdasher (approved):

Issue #308, opened 2023-05-01
Audit Premium Events SharePoint & Exchange


Sloane4:
# 💡 Summary #

Direct the user to enable 2 audit events through admin power shell to enable specific ability for Incident Response.
https://learn.microsoft.com/en-us/microsoft-365/compliance/audit-premium-setup?view=o365-worldwide#step-2-enable-audit-premium-events

## Motivation and context ##

Audit (Premium) features such as the ability to log crucial events such as MailItemsAccessed and Send require an appropriate E5 license assigned to users. Additionally, the Advanced Auditing app/service plan must be enabled for those users. You have to enable two Audit (Premium) events (SearchQueryInitiatedExchange and SearchQueryInitiatedSharePoint) to be logged when users perform searches in Exchange Online and SharePoint Online. In additional to the default policy that retains Azure AD, Exchange, OneDrive, and SharePoint audit records for one year, you can create additional audit log retention policies to meet the requirements of your organization's security operations, IT, and compliance teams. Once enabled you can search the audit log for Audit (Premium) events and other activities during forensic investigations of compromised accounts and other types of security or compliance investigations.

## Implementation notes ##

- Add new policy to Exchange baseline
- Add new policy to SharePoint baseline
- Create non implemented check in rego
- Update unit test cases for new policy
- Research if there is a way to check if audit events can be checked through SCuBA
- IF audit events can be checked open a new issue to enhance the code

## Acceptance criteria ##

- [ ] Updated policy has the change noted in implementation notes
- [ ] Non-Implementation check in Exchange & SharePoint
- [ ] All unit test cases pass

Sloane4:
I argue that this issue is dependent on the decision of #88. The previous issue asks if we should & the creation of an appropriate policy (should vs shall). While this issue focuses more on what to do if we add advanced audit logs (i.e. updating the baseline, code, testing). I would agree that this issue can be updated to better reflect that, but I see them as distinct issues.

tkol2022:
@Sloane4 @buidav  Can you please test the instructions you hyperlinked above to add the SearchQueryInitiatedSharePoint and SearchQueryInitiatedExchange audit event against a couple of our tenants that have the required license and then share your results here?

I'd like to make sure:

1. We flush out any nuances to the policy language.
2. We record the dependencies for these events to work. For example, it seems like a tenant user must have the MS 365 Advanced Auditing license turned on before an Administrator can enable the SearchQueryInitiated events noted above (see screenshot below)
[...]

PR #470, opened 2023-08-07
Formal internal review for Defender baseline

schrolla:
#  #
## 🗣 Description ##

This PR is designed to facilitate a formal review of the revised Defender baseline from CISA SCuBA principals and the Scuba development team. Following this review the baseline should be in a state for sending to the CISA agency formal review process.

      in code comments.
- [x] All new and existing markdown tests pass.

## ✅ Pre-merge checklist ##

- [x] Feature branch has been rebased against base branch with conflicts resolved

## ✅ Post-merge checklist ##

- [ ] Ensure Defender related markdown tests pass on merged branch

### DIFF
--- a/baselines/defender.md
+++ b/baselines/defender.md
@@ -388,8 +388,12 @@ information by restricted apps and unwanted bluetooth applications.
                access from restricted apps and unwanted bluetooth applications prevents unauthorized disclosure by those
                programs.
 - _Last modified:_ June 2023
-- _Note:_ The custom policy referenced here is the same policy
-          configured in MS.DEFENDER.4.1v1.
+- _Note:_
+  - The custom policy referenced here is the same policy
+    configured in MS.DEFENDER.4.1v1.
+  - This action can only be included if at least one device has been onboarded
+    to the agency tenant.  Otherwise, the option to block restricted apps will
+    not be available.
 
 ### Resources
 
@@ -404,6 +408,9 @@ information by restricted apps and unwanted bluetooth applications.
 
 - [Sensitive information \|
   NIST](https://csrc.nist.gov/glossary/term/sensitive_information)
+
+- [Get started with Endpoint data loss prevention - Microsoft Purview
+  (compliance) \| Microsoft Learn](https://learn.microsoft.com/en-us/purview/endpoint-dlp-getting-started?view=o365-worldwide)
 
 ### License Requirements
 
@@ -518,6 +525,10 @@ information by restricted apps and unwanted bluetooth applications.
    accessing protected files and to log the incident.
 
 #### MS.DEFENDER.4.6v1 instructions:
+
+If restricted app and unwanted Bluetooth app restrictions are desired,
+associated devices must be onboarded with Defender for Endpoint
+before the instructions below can be completed.
 
 1. Sign in to the **Microsoft Purview compliance portal**.
 
@@ -653,7 +664,7 @@ Microsoft Purview Audit (Standard) logging SHALL be enabled.
 - _Last modified:_ June 2023
 
 #### MS.DEFENDER.6.2v1
-Microsoft Purview Audit (Premium) logging SHALL be enabled.
+Microsoft Purview Audit (Premium) logging SHALL be enabled for ALL users.
 - _Rationale:_ Standard logging may not include relevant details necessary for visibility into user actions during an incident. Enabling Microsoft Purview Audit (Premium) captures additional event types that are not included with Standard. Furthermore, it is required for government agencies by OMB M-21-13 (referred to therein as by its former name, Unified Audit Logs w/ Advanced Features).
 - _Last modified:_ June 2023
 


baselines/defender.md
```diff
@@ -41,7 +41,7 @@ Microsoft Defender defines two [preset security
...
+Within the preset policies, users can be enrolled in [Exchange Online Protection](https://learn.microsoft.com/en-us/microsoft-365/security/office-365-security/eop-about?view=o365-worldwide), [Defender for Office 365 protection](https://learn.microsoft.com/en-us/microsoft-365/security/office-365-security/microsoft-defender-for-office-365-product-overview?view=o365-worldwide), and [Impersonation Protection](https://learn.microsoft.com/en-us/microsoft-365/security/office-365-security/anti-phishing-policies-about?view=o365-worldwide#impersonation-settings-in-anti-phishing-policies-in-microsoft-defender-for-office-365) can be configured.
```
crutchfield, line 44: Reads a little awkward for me.  Can I enroll in 3 or enroll in 2 and configure a 3rd?

baselines/defender.md
crutchfield, line 1: Extra line break between Rationale and Last Modified for MS.DEFENDER.1.1v1

crutchfield, line 1: Policy sections titles have mix of including colon in bold tag and not (e.g., _Note_:).

crutchfield, line 1: Section 3. Safe Attachments, second sentence, reads "do not have".  Should this read 'have"?

crutchfield, line 1: I know the custom policy referenced in 4.2v1 refers to the same one defined in 4.1v1 -- should this be more explicilty stated?

crutchfield, line 1: MS.DEFENDER.5.1v1 seems to either be duplicate of MS.EXO.16.1v1.  Perhaps I am missing a subtlety?

crutchfield, line 1: In MS.DEFENDER.5.2v1, the "or incorporated into a SIEM" phrase appropriate as an option to implement the policy.  Maybe better as a note - we are not providing instructions on how to incorporate SIEM tool.  And in general it would be difficult to check if used.  I think a useful addition in the future is an override capability.  We can say override MS.DEFENDER.5.2v1 with a rationale statement like "Using Splunk".  Then the report could pass that policy if the user 'explicitly" states they are using SIEM.

crutchfield, line 1: MS.DEFENDER.6.3v1 instructions state 1 year for retention policy but preface material in section 6 state 1 year hot storage and 18 months cold.  Also seems to indicate retention policy can naitively support this configuration.  I assume we would not be able to test if alternative "off-load" to another long term storage (say a 5.25 inch floppy disk) can be detected.   Maybe this option is a note and good candidate for future override with rationale.

crutchfield (changes requested):
Minor changes requested.  Several observation/questions that will not require action.

tkol2022:
### Sections 1 and 2 have policy dependencies

I noticed that policies 1.2, 1.3, 1.4, 1.5, 2.1, 2.2 and 2.3 seem dependent on 1.1 "The standard and strict preset security policies SHALL be enabled". From what I can tell, if you don't enable the present policies, the values that are set for the other policies noted above will not matter, because Defender won't enforce them.

If this is true:

- I would consider noting that dependency somehow in the policies. In Sharepoint MS.SHAREPOINT.1.3v1 and other policies, we have a "sorta" similar situation. I added a Note to those policies that says "This policy is only applicable if the external sharing slider on the admin page is set to any value other than Only People in your Organization."
- For ScubaGear, when evaluating policies 1.2, 1.3, etc. noted above, we might need to add code that checks whether 1.1 passed. If  the policy check for 1.1 fails, then I'm not sure it makes sense to provide a Pass for 1.2, 1.3 and all the dependent policies because that might give the user a false sense that they have implemented our policies correctly, when in fact, if they haven't enabled the preset policies, the system won't enforce 1.2, 1.3, etc.

image

tkol2022:
### MS.DEFENDER.1.1v1 instructions

The note after the step by step says "If the toggle slider is grayed out, click on Manage protection settings instead and configure the policy settings according to...some hyperlink...to enable each preset security policy." Why would the user need to go to that link? It seems all they have to do is follow steps 5 and 6 to turn on the preset security policies and in fact the hyperlink in the note provides the same instructions almost verbatim.

tkol2022:
### Clarity on the exception configurations

In policies MS.DEFENDER.1.2v1 and MS.DEFENDER.1.3v1 the note says "Specific user accounts, except for sensitive accounts, MAY be exempt from the preset policies, provided that they are added to one or more custom policies that offer comparable protection... Those users' accounts should be added to a custom policy that conforms as closely as possible to the settings used by the preset policies. "

- Not clear what is mean by "offer comparable protection"? If the protection is **comparable** (i.e. the same security protections), then why not just just keep those accounts in scope with the preset policies applied to everyone else? Are we saying that the custom policies can offer more or less security than the preset?
- Then it says "added to a custom policy that conforms as closely as possible to the settings used by the preset". This indicates that the custom policies are not really comparable - they are different - "as closely as".
- What is a real-world example of a situation where specific exempted user accounts would need either less or more security than the preset policies?
- Is ScubaGear going to be coded to support these exceptions?

tkol2022:
### MS.DEFENDER.1.4v1 and MS.DEFENDER.1.5v1 instructions

Optional steps 8 and 9 for these two policies direct the user to configure exception cases, but exceptions are actually described in the Notes field for policies 1.2 and 1.3. Policies 1.4 and 1.5 do not describe exception cases. Seems these steps should be in the instructions for 1.2 and 1.3. Also, logically policies 1.4 and 1.5 only configure the policies for "Specific recipients" so it doesn't seem to make sense on why you would need an exception in that case.

tkol2022:
### Transition from section 1 to section 2

I had a tough time transitioning from the scope of section 1 (preset policies) to section 2 (impersonation protection). When I look at all the policies in section 2, they all include "in both the standard and strict preset policies" and are configured at the same place where the section 1 policies are configured. To me it seems that section 2 is also about the preset policies and I'm not sure a separate section makes sense.

tkol2022:
### Safe documents - enhancement idea

This is a consideration for a future new policy. There is a setting named "Turn on Safe Documents for Office clients" configured on the same page as policy 3.1. Seems like a good feature to help manage risks in office clients. The documentation says that it uses the back-end version of Defender for Endpoint. Take a look and see if there is anything worth considering.

image

tkol2022:
### MS.DEFENDER.4.1v1 instructions

- Not sure if the portal has changed recently. The steps I followed to get to DLP > Policies were slightly different from steps 2 and 3. I got there from using the left navigation bar. Update the instructions to match the portal.

image

- Step 5 instructs the user to copy the existing "Default Office 365 DLP policy" instead of creating a new policy from scratch. This caused issues for me when I followed the steps below because when I clicked Submit to create the new policy, I received an error message related to some of the configurations that were copied from the "Default Office 365 DLP policy". Each tenant can have a "Default Office 365 DLP policy" in a different state because although it is named "Default" the policy can be changed. If you compare the default policies in the E5 versus G5 tenants (screenshots below) you will see that they have different rules. I'm wondering if it would be easier to have a user create a new policy from scratch instead of copying from the default.

image

image

image


- Step 7 - There is no Assign admin units option in the G5 tenant. That is currently a preview feature. I do see it in the E5 tenant.

- Step 8 is missing ", then click Next".

- For step 8, I set Status to On for all products except Power BI and Microsoft Defender for Cloud Apps but that caused an error message because On-premises repositories was selected. I had to unselect that to get rid of the error message and continue to the next screen. Here is a screenshot of the error:

image

- When I clicked Submit the system displayed an error message. Apparently the names in the Policy settings are too long by default - they need to be 64 characters or less. You might need to update the instructions to reflect this because the error message is cryptic and I sort of figured it out by guessing.

image

image

tkol2022:
### MS.DEFENDER.4.3v1 The action for the DLP policy SHOULD be set to block sharing sensitive information with everyone when DLP conditions are met.

- Wondering if we can remove the "when DLP conditions are met" from the policy title to make it more concise?

tkol2022:
### MS.DEFENDER.4.5v1 A list of apps that are restricted when accessing files protected by DLP policy SHOULD be defined

- I recommend changing the language to "be enforced" or something like that since defined makes it sound like you are just documenting the restricted apps.
- Consider if there is any way to make this policy more specific or if there is a way we can provide agencies with guidance on a standard list of apps they can restrict. Right now, the burden is on the agency to define which endpoint apps should be restricted, and that seems very open ended to me since there are thousands of apps that can run on an endpoint. We had a similar situation in the Teams baseline for one of the app settings and Grant had dropped a comment similar to this for that policy.

tkol2022:
### MS.DEFENDER.4.2v1

- Nitpick. Recommend changing the policy text from "applied in" to "applied to". Applied in makes it sound like you have to go to each tool and configure it within separate admin portals.

tkol2022:
### Combining policies in Section 4 Data Loss Prevention

- After analyzing this section and consulting with @schrolla I came to the understanding that section 4 contains an **anchor** policy 4.1 and that other policies (4.2, 4.3, 4.4) are all dependent on 4.1. Upon further inspection when going through the instructions, to me, it seems that if you follow the implementation instructions described for 4.1 you are already covering the configuration items described in policies 4.2, 4.3 and 4.4, so it wasn't clear to me if these three policies are needed? In theory, if I configure 4.1 as described in its instructions, I would be compliant with 4.2, 4.3 and 4.4? If that is true, then why do we need these separate policies.
- If it turns out that we in fact do have a reason for 4.2, 4.3 and 4.4 being their own policies, then I think it might be worth describing that they are dependent on 4.1. This is similar to my earlier comments on section 1 and 2 dependencies. I think documenting dependencies servers two purposes. First it helps the user that is reading these baselines how policies are interconnected and affect each other and second it provides documentation that ScubaGear developers can follow to properly code the dependencies in the code. We had bugs in the past with Sharepoint / OneDrive where code implementations were missing dependencies and generating incorrect policy reports.
- Policy 4.6 also depends on 4.1.
- The policy names in section 4 starting with policy 4.2 contain language such as "The custom policy" and "The action for the DLP policy". It is not obvious what "the" is referring to in these instances. I'm pretty sure "the" refers to the custom policy for 4.1 but it is not obvious. Consider rephrasing for specificity.

tkol2022:
### MS.DEFENDER.4.6v1 instructions

- Step 7 Under Actions, click Add an action. The add action button was greyed out for me so I wasn't able to complete the steps.

image

tkol2022:
### MS.DEFENDER.5.1v1 instructions

- Steps 4 and 5 instruct the user to select the policy name and turn the status to On. Which policy is being referred to here? If this is meant to refer to all of the policies, then we should change the instructions. In the E5 tenant there are 48 alert policies. If you select all of them and then click on the Edit at the top of the page, it gives you the option to "Enable selected policies" which I'm guessing would turn their status to On? Maybe we can revise the instructions to reflect this.

tkol2022:
### MS.DEFENDER.5.2v1 The alerts SHOULD be sent to a monitored address or incorporated into a SIEM

- We mention that the alerts can be incorporated into a SIEM. If Microsoft has any reference pages on sending Defender alerts to a SIEM, I recommend we include them in the resources section.
- Also in the MS.DEFENDER.5.2v1 instructions section I couldn't find a way to configure the alert email recipients for all 48 alert policies in bulk. This indicates that the user might have to configure each one individually. Seems like a lot of work. I suppose if there's a way to send these to the agency SIEM then this configuration isn't necessary.

tkol2022:
### MS.DEFENDER.6.2v1 Microsoft Purview Audit (Premium) logging SHALL be enabled

- From my read of the policy and the instructions it is not clear if we are recommending that the admin turn on premium logging for ALL users in the tenant or a select set of users (e.g. high risk espionage targets)? Since it is not specific one could assume we are recommending that the premium logging be applied to all users. I'm thinking that we might want to add a recommendation to further clarify what the policy and implementation actually are.
- If the policy is in fact to turn on premium logging for all users, I don't think it is feasible to follow Microsoft's instructions and perform the configuration on a user by user basis. Even for small tenants this would be cumbersome and prone to error, but for large tenants this would not be practically feasible. I'm wondering if there is a scripted way to apply premium logging or some other alternative. I realize this may not be our problem to solve :)

tkol2022:
### Reminder about related issue

This is just a reminder that we had a related issue #308 assigned to this release which is about enabling some additional premium events that don't get logged by default when you enable premium logging for a set of users. Not sure if the events in that issue would go in Defender or EXO.

tkol2022:
MS.DEFENDER.6.3v1 Audit logs SHALL be maintained for at least the minimum duration dictated by OMB M-21-31

- If the agency chooses to retain their audit logs in their centralized SIEM or other solution, instead of within M365, then does this policy still apply? I think we should mention that in the policy and in the instructions. Or do we want the agency to retain their logs within M365 for the required duration (even if the SIEM is the "master" system) as a resiliency measure? Might be worth it.

- Regarding the instructions section, Microsoft's audit retention policy configuration instructions do not provide specific guidance on two aspects of the configuration that we do not mention in our instructions either: the Users and the Record type. You must configure at least one of those two. Are we expecting the agency to keep the Users field blank? Per Microsoft that indicates that the policy will apply to all users? Sounds logical to me. If that is the case, then the user must select a set of Record types that the retention applies to. Are we expecting the user to select all of the available record types? There are dozens of them. Whatever the case may be, I think we need to add these details to help direct the user a little more in our instructions.

tkol2022 (approved):
I'm approving this PR with the assumption that comments will be addressed. Thanks for letting me provide feedback. Nice work.

crutchfield (approved):
Thanks for clarifications on previous comments.  Looks good.

Issue #164, opened 2023-02-21
Agency 4 Pilot: Exchange 2.2 false negative


ahuynhMITRE:
Agency 4 noted that they comply with the following policy "An SPF policy(s) that designates only these addresses as approved senders SHALL be published" but the report is showing "FAIL". Further investigation into SPF and DNS records is already underway so this may be linked.

ethanb-cisa:
@gatesry one of our theories is this is because looking up a domain internally using an internal nameserver will return different results than looking it up externally in some cases. See [this comment](https://github.com/cisagov/ScubaGear/issues/41#issuecomment-1411086060) for more.

Do you know if this could be happening for your org? We had an idea to use DoH but that may cause other issues (proxy, etc).

gatesry:
@ethanb-cisa We just spun up an off-domain VM; this allowed it to grab the correct SPF records.

tkol2022:
### Reviewed 12/14
@ahuynhMITRE @adhilto Please review and close if this is OBE.
If it is still valid, please assign a size label.
I'd like to see more details in the description.

adhilto:
Addressed by adding a DoH fallback, closing.

PR #515, opened 2023-09-01
GET dns txt records from https://dnsviz.net/d/<domain_name>/responses/ as a fallback for unsuccessful dns resolutions.

c0rgifanclub:
#  #

## 🗣 Description ##

Provides another means of getting dns txt records for spf/dkim/dmarc checks, in the environments where dns resolution traffic is restricted.

To address https://github.com/cisagov/ScubaGear/issues/164, where DNS split horizon and possible network blocks prevent access to necessary dns TXT records.

Uncomment this section if a screenshot is needed.
image

- [ ] *All* future TODOs are captured in issues, which are referenced
      in code comments.
- [ ] All relevant type-of-change labels have been added.
- [ ] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [ ] All relevant repo and/or project documentation has been updated

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Create a release.

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
@@ -264,6 +264,36 @@ function Invoke-RobustDnsTxt {
         }
     }
 
+    if (-not $Success) {
+        # Traditional DNS and DoH queries failed. Try using the dnsviz.net server provided by DNS-OARC
+        # (https://dnsviz.net/d/$Qname/dnssec/).
+        while ($TryNumber -lt $MaxTries) {
+            $TryNumber += 1
+            try {
+                $dnsVizResponse = Invoke-WebRequest -Uri "https://dnsviz.net/d/$($Qname)/responses/" -ErrorAction Stop
+                if ($dnsVizResponse.StatusCode -eq 200) {
+                    # http status code 200 indicates there was no error
+                    # for each tr element, get the td elements with class name "rr"
+                    $dnsVizResponse.ParsedHtml.getElementsByTagName("tr") | ForEach-Object { if ($_.outerhtml -like "*TXT*") { $_.getElementsByClassName("rr") | foreach-Object {$Answers += $_.innerText}} }
+                    $LogEntries += @{"query_name"=$Qname; "query_method"="DNSViz.net_WebService"; "query_result"="Query returned $($textrecords.Length) txt records"}
+                    $Success = $true
+                    write-output "Query returned response code $($dnsVizResponse.Status)"
+                    break
+                }
+                else {
+                # The remainder of the response codes indicate that the web service request did not succeed.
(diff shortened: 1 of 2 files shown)


adhilto:
Thanks @c0rgifanclub for this contribution! I've actually worked with DNSViz in the past, it's a great tool. None of this should be news for you, but I'm going to add some context about DNSViz for the other developers. I also have some considerations to share to get the conversation going, discussing some pros and cons of adding this.

This pull request aims to tighten a current visibility gap we have with ScubaGear. Specifically, we noticed during the pilots that for some organizations, `Resolve-DnsName` would not return any answers for their domains, even though DNS records were published for those domains and could be retrieved when queried external from their network. We've mitigated this by adding a fallback to DNS over HTTPS (DoH). In most cases, this fixes the situation, but some organizations additionally block DoH. This pull request adds a third fallback, namely a web request to [DNSViz](https://dnsviz.net/). DNSViz is a tool, originally developed at Sandia National Labs, designed to allow administrators visualize the DNSSEC chain of trust. Part of DNSViz's output includes a section with the txt records for the domain. The PR adds code which makes a web request to DNSViz and parses the HTML response to extract the txt records.

The main benefit of this PR is that it further restricts our DNS visibility gap and allows users a path forward if their traditional DNS request fails and DoH is also blocked. Other considerations:
- Adds latency. I haven't tested it myself, but per @c0rgifanclub's comment, we can expect 9 seconds of latency (I'm assuming per domain, which could add up). Though this would be the 3rd fallback, so hopefully most users wouldn't see this latency.
- In most cases where DNSViz is needed, manual intervention will still be required. If DNSViz hasn't see a given domain before, you'll need to manually open the website and trigger an analysis, wait a few minutes, then re-run ScubaGear. That said, this at least gives users a path forward. Our current implementation just displays a warning if applicable but doesn't indicate a way for the users to resolve the warning.
- The code does not use an API, but rather parses the HTML, meaning it will break if the web interface is significantly changed. That said, I would consider the web interface to be very stable. I anticipate that any instances of this happening would be rare.

@c0rgifanclub feel free to weigh in on any of the above points if you'd like. We'll then discuss this internally to determine the best path forward.

adhilto:
Thank you again for your contribution, but we will be closing this PR. Ultimately, we elected to not add this for the following reasons:
- Java dependency. https://dnsviz.net/ actually uses Java in the browser. Which can result in some fun PowerShell issues and overly complicates ScubaGear's dependencies.
- Latency. Assuming you got it working, it would add significant latency, about 9 seconds per domain, representing an obvious scalability concern.
- User experience. In most cases where DNSViz is needed, manual intervention will still be required. If DNSViz hasn't see a given domain before, you'll need to manually open the website and trigger an analysis, wait a few minutes, then re-run ScubaGear, a particularly user-costly workflow.

Once again, thank you for your contribution.

PR #597, opened 2023-10-13
Fix MS.TEAMS.1.5 Policy of ScubaGear

### LINKED ISSUE #452: Implement the new Teams policy MS.TEAMS.1.5v1 in Rego
# 🐛 Summary #

Teams baseline policies are being regrouped and reorganized. As part of this, a new sub policy has been created - the check for which is not implemented in the previous versions of ScubaGear. The policy is MS.TEAMS.1.5v1 (Dial-in users, SHOULD NOT be enabled to bypass the lobby). This sub policy is closely related to MS.TEAMS.1.3v1 and can be accomplished in one of two configurations - the earlier releases have been covering one of the config but not the other config check. New rego script has to be developed to explicitly cover the incomplete config check. 

## To reproduce ##

1. Sign in to the Microsoft Teams admin center. 
2. Select Meetings > Meeting policies.
3. Select the Global (Org-wide default) policy and Custom Policies
4. Under the Meeting join & lobby section, set 'Who can bypass the lobby' to 'People in my Org' or one of the other
5. Then set the 'People dialing in can bypass the lobby' to Off. <-- This is currently not explicitly checked and it is required per policy. 

## Expected behavior ##

The result for the config 'People dialing in can bypass the lobby' to Off should be called out in the report based on the config quoted in the reproduce steps.  

Add any screenshots of the problem here.

nanda-katikaneni:
#  #

## 🗣 Description ##
Currently ScubaGear (in Emerald) shows incorrect results for the MS.TEAMS.1.5v1 - with Criticality and Details columns of report showing undeclared values. The objective of this policy is to check that dial-in users are not configured to be auto admitted into meetings and it is applicable to all policies. The 'who can bypass the lobby' setting tested through another teams policy (1.3) auto selects the true/false value for the dial-in users. But there are some configuration paths that can leave this setting (dial-in users to bypass lobby) to be in true mode. This policy enforces and checks this setting to be false irrespective of previous setting and for all meeting policies. So the implementation is: for all meeting policies, check that 'dial-in users can bypass the lobby' is set to false and count any policies that are non-compliant. 

closes #452 

## 🧪 Testing ##
Tested the feature branch with G5 and E5 tenants with following test cases - all passed:

In Teams -> Meeting Policies -> Meeting Join and Lobby, do the following:
1. Set "Who can bypass the lobby" to one of: People in my org, People in my org and guests, People in my org, trusted org, and guests.
2. Set "People dialingin can bypass the lobby" to Off

Above 3 tests should pass the policy 1.5. For negative tests,  set "People dialingin can bypass the lobby" to On and make sure that policy 1.5 fails with a detail that "either everyone or dial-in users are allowed to bypass the lobby". Test this for a custom policy as this policy is applicable to all policies. 

- [ ] *All* future TODOs are captured in issues, which are referenced
      in code comments.
- [ ] All relevant type-of-change labels have been added.
- [ ] All relevant repo and/or project documentation has been updated

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Create a release.


Rego/TeamsConfig.rego
```diff
@@ -150,37 +150,26 @@ tests[{
...
+	"ActualValue" : Policies,
+	"ReportDetails" : ReportDetailsArray(Status, Policies, String),
 	"RequirementMet" : Status
 }] {
-	Policy := input.meeting_policies[_]
-	# This control applies to all policies including custom, so no filter
-	Conditions := [Policy.AutoAdmittedUsers != "Everyone", Policy.AllowPSTNUsersToBypassLobby == false]
-    Status := count([Condition | Condition = Conditions[_]; Condition == false]) == 0
+	Policies := MeetingsAllowingPSTNBypass
+	String := "meeting policy(ies) found that either allow everyone or dial-in users to bypass lobby:"
```
crutchfield, line 168: I guess the statement is still true.  But it through me off since you are not longer explicitly checking foe "everyone".

crutchfield (approved):
Looks good.  I did make an observation but it does not require action.

rgbrow1949:
Checked to see if the test would fail if I changed the global policy so that "People dialing in can bypass the lobby" was set to on. 1.3 failed, but 1.5 still passed unexpectedly. 
image

nanda-katikaneni:
> Checked to see if the test would fail if I changed the global policy so that "People dialing in can bypass the lobby" was set to on. 1.3 failed, but 1.5 still passed unexpectedly. image

Are you sure the test rant with the branch? - I ask because the result in your test for 1.5 shows criticality and details as 'system.object'. That is fixed with the branch and in my test against G5 tenant shows the following result - note the criticality and details columns against 1.5.
Screenshot 2023-10-16 at 7 15 08 PM

rgbrow1949 (approved):
Tested this with correct configuration and incorrect configuration. The new code works as intended for both states. Approved.

PR #651, opened 2023-11-08
Fix versioned tag in URLs

### LINKED ISSUE #650: ScubaGear URL version inconsistent with tags
# 🐛 Summary #

The `ScubaBaseURL` used in assessment report details and `MarkdownLink` in CreateReport.psm1 do not generate versioned URLs correctly.  As a result, links in Rego and HTML reports will be broken and prevent users from being able to reference the proper versioned baseline associated with a report.

## To reproduce ##

1. Run the latest version of the ScubaGear tool
2. Open the HTML report and choose an individual product report
3. Find a baseline URL reference and click on it.  Note the URL contains the version '1.0.0'.
4. Notice that in form of version tags in the repo is 'vX.Y.Z'.

## Expected behavior ##

The expectation is that the generated URL includes the tag with the 'v' prefix to match the actual URL at which the versioned baseline can be found.

schrolla:
#  #

## 🗣 Description ##

This request updates both the Rego report utility `ScubaBaseURL` variable and PowerShell HTML report variable `MarkdownLink` in `CreateReport.psm1` to generate versioned URLs to GitHub using the 'v' prefix in front of the release version of the tool.

Without this fix and with the existing build and sign release action, ScubaGear version tags include the 'v' prefix.  However, URLs in Rego and the HTML report are built without the prefix.  As a result, URLs will be broken and incorrect and users will not be directed to the correct website location for baseline references in test results and reports.

Closes #650 

## 🧪 Testing ##

Ran `Invoke-Scuba -p *` to generate a full set of reports and validated that URL references to the baselines in report details contains the proper 'v' prefix.

Generated HTML report with URL preview shows report for MS.AAD.6.1v1 correctly generated a URL link with version as `v1.0.0`.

Screenshot 2023-11-08 at 4 35 56 PM

      in code comments.

## ✅ Pre-merge checklist ##

- [ ] Rebase branch against parent branch as needed prior to merge

## ✅ Post-merge checklist ##

- [ ] Validate parent branch tests still pass

### DIFF
--- a/PowerShell/ScubaGear/Modules/CreateReport/CreateReport.psm1
+++ b/PowerShell/ScubaGear/Modules/CreateReport/CreateReport.psm1
@@ -168,7 +168,7 @@ function New-Report {
         $Number = $BaselineName.ToUpper() + '-' + $BaselineGroup.GroupNumber
         $Name = $BaselineGroup.GroupName
         $GroupAnchor = New-MarkdownAnchor -GroupNumber $BaselineGroup.GroupNumber -GroupName $BaselineGroup.GroupName
-        $MarkdownLink = "<a class='control_group' href=`"$($ScubaGitHubUrl)/blob/$($SettingsExport.module_version)/baselines/$($BaselineName.ToLower()).md#$GroupAnchor`" target=`"_blank`">$Name</a>"
+        $MarkdownLink = "<a class='control_group' href=`"$($ScubaGitHubUrl)/blob/v$($SettingsExport.module_version)/baselines/$($BaselineName.ToLower()).md#$GroupAnchor`" target=`"_blank`">$Name</a>"
         $Fragments += $Fragment | ConvertTo-Html -PreContent "<h2>$Number $MarkdownLink</h2>" -Fragment
     }
 
--- a/Rego/Utils/ReportUtils.rego
+++ b/Rego/Utils/ReportUtils.rego
@@ -7,7 +7,7 @@ default BaselineVersion := "main"
 BaselineVersion := input.module_version
 
 # baselineVersion := "3.0.0." # Baseline version is pinned to a module version
-ScubaBaseUrl := sprintf("https://github.com/cisagov/ScubaGear/blob/%v/baselines/", [BaselineVersion])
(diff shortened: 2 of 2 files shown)


crutchfield (approved):
Verified navigation to v0.2.1 with changes.

adhilto:
Just one question, do we want the link at the top right to point to the versioned baselines? Currently it just points to main.
image

I think pointing to the versioned baselines might be better, but I don't have strong feelings about it. Everything else looks good.

adhilto:
> > Just one question, do we want the link at the top right to point to the versioned baselines? Currently it just points to main.
> > I think pointing to the versioned baselines might be better, but I don't have strong feelings about it. Everything else looks good.
> 
> That is a very good catch. I think you're right and that it should point to the versioned baseline. @crutchfield what do you think?
> 
> Although on second look, this is hardcoded into both the parent and individual templates right now to the main branch. So the fix gets a bit more involved, but changing the New-Report method to search for a new {BASELINE_URL} template and doing the search and replace. It isn't terribly difficult, but definitely more involved than the previously proposed change. So we may want to put this in as a bug for future fixes instead.

I'm good saving it for a future change. Either way, I'll go ahead and approve.

adhilto (approved):

crutchfield:
> > > Just one question, do we want the link at the top right to point to the versioned baselines? Currently it just points to main.
> > > I think pointing to the versioned baselines might be better, but I don't have strong feelings about it. Everything else looks good.
> > 
> > 
> > That is a very good catch. I think you're right and that it should point to the versioned baseline. @crutchfield what do you think?
> > Although on second look, this is hardcoded into both the parent and individual templates right now to the main branch. So the fix gets a bit more involved, but changing the New-Report method to search for a new {BASELINE_URL} template and doing the search and replace. It isn't terribly difficult, but definitely more involved than the previously proposed change. So we may want to put this in as a bug for future fixes instead.
> 
> I'm good saving it for a future change. Either way, I'll go ahead and approve. I can re-review if we end up wanting the change now after all.

I think it should always point to the released/tagged version.  But agree we should look at it in next release.

PR #769, opened 2023-12-28
Consolidate redundant PowerShell linting workflows

james-garriss:
## 🗣 Description ##

After adding the MegaLinter (ML) PowerShell linter, we had two linters for PowerShell, both of which used PS Script Analyzer (PSA).  The original workflow was not working; that is, it never returned any linting issues.  The newer one that used ML was working fine.  The original was based upon code created by an [individual](https://github.com/devblackops) and has not been updated for a couple years.  I tested with both the [GitHub Action and Docker versions](https://github.com/devblackops/github-action-psscriptanalyzer), but neither worked.  In the end, I removed the old workflow and kept the new one.   Using ML allows us to add as many different linters as we want using the same tool and nearly the same workflows.

There is no reason to have two workflows that use the same linter for PS.

## 🧪 Testing ##

I tested by pushing code and viewing the results in Actions.

## ✅ Pre-merge checklist ##

- [ ] Finalize version.

## ✅ Post-merge checklist ##

- [ ] Create a release.


james-garriss:
To @crutchfield and @dagarwal-mitre : This action is currently failing because some of the code has PS linting issues.  These should be resolved by the owners of the code, as we agreed when we added the ML linter workflows for the first time.

crutchfield (changes requested):
Can you stop redundant runs?  This happens on several actions. We need a pattern to eliminate.  

image

dagarwal-mitre (approved):

james-garriss:
> Can you stop redundant runs?

Removed pull request trigger, per our conversation.

.github/workflows/run_linter_powershell.yaml
```diff
@@ -7,13 +6,13 @@ name: Lint Check for PowerShell
...
   push:
     paths:
-      - '**.ps1'
-      - '**.psm1'
-      - '**.psd1'
-      - '**.pssc'
-      - '**.psrc'
-      - '**.ps1xml'
-      - '**.cdxml'
+      - "**.ps1"
```
crutchfield, line 9: What is rationale for changing from single to double quotes?

crutchfield (approved):
Looks good.  All annotations not directly related to this PR.

.github/workflows/run_linter_powershell.yaml
james-garriss: This is a result of using linters on the project.  While there are many ways to represent strings, YAML linters generally consider double-quotes to be safer because it allows escaping characters in an expected manner.  If we don't want to use this check, I can add a new issue to update the YAML linter accordingly.  I don't have dog in this fight personally.

james-garriss:
@nanda-katikaneni , I think this is ready to merge into flipper.  Thank you!

PR #893, opened 2024-02-06
Fixes ScubaConfig.LoadConfig repeated invocations

### LINKED ISSUE #844: ScubaConfig.LoadConfiguration only works the first time
# 🐛 Summary #

Calling `ScubaConfig.LoadConfiguration` once works; calling it again after than does not load the new config file. In both cases the output of `LoadConfiguration` is `true` even though no config is loaded in the second. `LoadConfiguration` checks if `[ScubaConfig]::_IsLoaded` and only loads the new config if that flag is false (ie., after `ResetInstance` or no configuration has been loaded yet). It should either throw an error or just overwrite the existing config.

Other random thing.  `hidden [Object]$Configuration` should be `hidden [hastable]$Configuration` since the output of CovertFrom-Yaml is a hashtable. Apparently PS class type declarations are actually not checked at runtime, but it would help avoid confusion.  

## To reproduce ##
config-test1.json:
```
{
    "ProductNames": [
      "teams"
    ]
}
```

config-test2.json:
```
{
    "ProductNames": [
      "exo"
    ]
}
```

test.ps1:
```
using module ./PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1

[ScubaConfig]::ResetInstance()
$cfg = [ScubaConfig]::GetInstance()
write-host "Loading config 1"
$loaded = $cfg.LoadConfig("./config_test1.json")
write-host "loaded: " $loaded
write-host "ProductNames1: " $cfg.Configuration.ProductNames
write-host "Loading config 2"
$loaded = $cfg.LoadConfig("./config_test2.json")
write-host "loaded: " $loaded
write-host "ProductNames2: " $cfg.Configuration.ProductNames
```
The output of this script is:
```
Loading config 1
loaded:  True
ProductNames1:  teams
Loading config 2
loaded:  True
ProductNames2:  teams
```
## Expected behavior ##

If I load `config-test1.json` it should reflect the values in the file. If I then load `config-test2.json` it should reflect the values in the. new file. This could be easily solved by calling `.ResetInstance` from inside `.LoadConfiguration`

Happy to take this one if it would be helpful.

### LINKED ISSUE #845: Add docs on how to use ScubaConfig
# 🐛 Summary #

`ScubaConfig` is designed to function as a singleton, such that only a single instance of the class exists at any given time. Calling `[scubaconfig]::GetInstance()` repeatedly will always return the same instance. This has some potentially confusing and unexpected effects, such as the loaded configuration persisting for the life of the powershell process, even from within in other modules that subsequently import the ScubaConfig module, possibly expecting a newly created config object to be empty. It requires client code to aggressively manage the state of the ScubaConfig cached instance to avoid inadvertently pulling in configs from a prior usage. This has implications for tests, where its necessary to call `[scubaconfig]::ResetInstance()` before and after tests to avoid pulling in configs from a prior test or accidentally injecting configs into a subsequent test.

There was most likely a compelling reason to make this a singleton, but I think the benefits are outweighed by the potentially confusing results. It would be more idiomatic and straight forward to make ScubaConfig a normal instance that people create and throw away when it's done.

There is a question of how to do this without having to change everything. I would suggest keeping `GetInstance` as a staticmethod, but turn it into a normal constructor that accepts the path to the config file as a positional arg. It would load the config and return a new instance of a scubaconfig. If no config is given, to preserve backward compatibility maybe it could look for an environment variable SCUBA_CONFIG by default and load the file from that path. If none of above, maybe throw an error.

If the motivation was to make it easy to find the loaded `scubaconfig` without having to manually pass it around everywhere, I would probably suggest we call a spade a spade and just explicitly make it a global variable. Both these methods result in the config lingering around the shell session permanently after Scuba runs unless affirmatively removed, so cleanup actions are necessary with both.

## To reproduce ##

1. Run some tests: `invoke-pester -path ./PowerShell/ScubaGear/Testing/Unit/PowerShell/ScubaConfig/`
2. Import scubaconfig: `using module ./PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1` 
3. The last configuration loaded during the tests is still in place: `[scubaconfig]::GetInstance().Configuration`

## Expected behavior ##

I expected `[scubaconfig]::GetInstance()` to return an empty config the first time after importing the scubaconfig module. It will return whatever the last used scubaconfig object was. The orchestrator avoids any problems by removing and reimporting the scubaconfig module before running  👍 and calling `[scubaconfig]::ResetInstance()` after running. This has to be done carefully in any tests that incorporate ScubaConfig so as to avoid state persisting between runs and unexpected things happening.

Output:
```
> [scubaconfig]::GetInstance().Configuration
Name                           Value
----                           -----
OutProviderFileName            ProviderSettingsExport
M365Environment                commercial
OutReportName                  BaselineReports
ProductNames                   {aad, defender, exo, onedrive…}
.
DisconnectOnExit               False
AnObject                       {[name, MyObjectName]}
OutRegoFileName                TestResults
Description                    JSON Configuration file for unit test in SCuBA
OPAPath                        /home/tneale/feat/pr782/PowerShell/ScubaGear/Modules/ScubaConfig/../../..
OutPath                        .
OutFolderName                  M365BaselineConformance
LogIn                          True
```

Happy to take this one if it would be helpful.

twneale:
# Fixes ScubaConfig.LoadConfig repeated invocations #
  

## 🗣 `.LoadConfig` now always calls `.ResetInstance` before loading config ##

This PR fixes #844, in which calling `.LoadConfig` more than once has no effect. The fix is trivial and just invokes `.ResetInstance` before loading a new file.

Primarily needed to avoid surprises during testing.

Also closes #845, added a comment-based help docstring to ScubaConfig, per discussion with team; simply easier to do that one here as well.

## 🧪 Testing ##

Added a unit test almost identical to the diagnostic test outlined in #844 

## ✅ Pre-merge checklist ##

- [x] Notified merge coordinator that PR is ready for merge via comment mention

## ✅ Post-merge checklist ##

### DIFF
--- a/PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1
+++ b/PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1
@@ -22,13 +22,12 @@ class ScubaConfig {
         if (-Not (Test-Path -PathType Leaf $Path)){
             throw [System.IO.FileNotFoundException]"Failed to load: $Path"
         }
-        elseif ($false -eq [ScubaConfig]::_IsLoaded){
-            $Content = Get-Content -Raw -Path $Path
-            $this.Configuration = $Content | ConvertFrom-Yaml
+        [ScubaConfig]::ResetInstance()
+        $Content = Get-Content -Raw -Path $Path
+        $this.Configuration = $Content | ConvertFrom-Yaml
 
-            $this.SetParameterDefaults()
-            [ScubaConfig]::_IsLoaded = $true
-        }
+        $this.SetParameterDefaults()
+        [ScubaConfig]::_IsLoaded = $true
 
         return [ScubaConfig]::_IsLoaded
     }
--- a/PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1
+++ b/PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1
@@ -38,7 +38,7 @@ class ScubaConfig {
     }
 
     hidden [Guid]$Uuid = [Guid]::NewGuid()
-    hidden [Object]$Configuration
+    hidden [hashtable]$Configuration
 
     hidden [void]SetParameterDefaults(){
         if (-Not $this.Configuration.ProductNames){
--- a/PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1
+++ b/PowerShell/ScubaGear/Modules/ScubaConfig/ScubaConfig.psm1
@@ -1,4 +1,20 @@
 class ScubaConfig {
+    <#
+    .SYNOPSIS
+      This class stores Scuba config data loaded from a file.
+    .DESCRIPTION
+      This class is designed to function as a singleton. The singleton instance
+      is cached on the ScubaConfig type itself. In the context of tests, it may be
+      important to call `.ResetInstance` before and after tests as needed to
+      ensure any preexisting configs are not inadvertantly used for the test,
+      or left in place after the test is finished. The singleton will persist
+      for the life of the powershell session unless the ScubaConfig module is
+      removed. Note that `.LoadConfig` internally calls `.ResetInstance` to avoid
+      issues.
+    .EXAMPLE
+      $Config = [ScubaConfig]::GetInstance()
+      [ScubaConfig]::LoadConfig($SomePath)
+    #>
     hidden static [ScubaConfig]$_Instance = [ScubaConfig]::new()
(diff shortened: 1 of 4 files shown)


twneale:
@schrolla This one is super simple and is ready to rip as soon as the approvals come through cc @crutchfield @Dylan-MITRE

crutchfield (approved):
Looks good

Dylan-MITRE (approved):
Screenshot 2024-02-07 at 8 02 56 AM

Looks good, actually reflect the values in files.

PR #1094, opened 2024-05-10
Improve AAD provider performance by directly invoking Graph APIs for Cmdlets with slow load times

### LINKED ISSUE #816: prototype a performance redesign of the AAD provider
# 💡 Summary #

Perform hands-on prototyping with multi-threading and other techniques to determine if it is possible to significantly improve the performance of the AAD provider with a redesign of specific code blocks.

## Motivation and context ##

Currently the AAD provider is relatively slow because it has to fetch information about users and groups that are assigned to privileged roles and the current process invokes MS Graph API calls several times within multiple loops.

## Implementation notes ##

Develop and test various alternative implementation techniques for the Get-PrivilegedUser and Get-PrivilegedRole functions. Perform profiling to compare executions times of alternative implementations versus the current code. Keep profiling data in a spreadsheet for reference.

Some potential alternative implementation technique that may be considered are listed below. This is not an exhaustive list - the assignee can experiment with techniques not listed here.

- Using Powershell RunspacePool to perform concurrent Graph API requests
- Using [MS Graph JSON batching](https://learn.microsoft.com/en-us/graph/json-batching)
- Replacing the existing Powershell Cmdlets with direct calls to Graph API via Invoke-MgGraphRequest
- any ideas not covered here

## Acceptance criteria ##

- [ ] All techniques have been developed and tested with their profiling data collected

tkol2022:
## 🗣 Description ##

This code enhancement of the AAD provider improves the performance of calls to Powershell cmdlets that are part of the Microsoft.Graph.Beta.Identity.Governance module. The enhancement replaces calls to those cmdlets with direct calls to the respective MS Graph REST APIs. The reason this works to cut down the execution time is that we found that the loading of the Identity.Governance module into memory when ScubaGear first runs in a fresh Powershell instance, is commonly a slow activity, in some cases taking over a minute just to load. By calling MS Graph APIs directly we bypass the penalty associated with loading the module into memory. Based on limited testing running in a virtual machine this enhancement resulted in a 42% reduction in execution time against our E5 tenant and a 35% reduction against our G5 tenant.

closes #816 

The AAD provider takes considerably longer to run compared to the other providers and in general runs slowly sometimes taking several minutes to complete. This creates delays for our development team who frequently run the code and have to perform testing and it slows down the automated processes that execute the code as well. Although we haven't received requests from end users to improve the performance, a positive side benefit is that their experience will improve when using the tool as well.

## 🧪 Testing ##

These code changes were tested against the E5, G5, G3 and GCC High tenants.

## ✅ Pre-merge checklist ##

## ✅ Post-merge checklist ##

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
@@ -18,11 +18,28 @@ $GraphEndpoints = @{
 
 function Invoke-GraphDirectly {
     param (
-      [Parameter(Mandatory)][string]$commandlet,
-      [System.Collections.Hashtable]$queryParams
+        [ValidateNotNullOrEmpty()]
+        [string]
+        $commandlet,
+
+        [ValidateNotNullOrEmpty()]
+        [string]
+        $M365Environment,
+
+        [System.Collections.Hashtable]
+        $queryParams
     )
+
     Write-Debug "Replacing Cmdlet: $commandlet"
     $endpoint = $GraphEndpoints[$commandlet]
+    # GCC high and DoD have slightly different endpoint URLs
+    if (($M365Environment -eq "gcchigh") -or ($M365Environment -eq "dod")) {
+        $endpoint = $endpoint.Replace(".com", ".us")
+    }
+    if ($M365Environment -eq "dod") {
+        $endpoint = $endpoint.Replace("graph", "dod-graph")
+    }
+
     if ($queryParams) {
         <#
           If query params are passed in, we need to augment the endpoint URI to include the params. Paperwork below. 
@@ -38,6 +55,7 @@ function Invoke-GraphDirectly {
         $endpoint = $uri.ToString()
     }
     Write-Debug "Graph Api direct: $endpoint"
+    
     # try {
         $resp = Invoke-MgGraphRequest -Uri $endpoint -UserAgent 'ScubaGear'
     # }
@@ -56,6 +74,13 @@ function Export-AADProvider {
     .Functionality
     Internal
     #>
+
+    [CmdletBinding()]
+    param (
+        [ValidateNotNullOrEmpty()]
+        [string]
+        $M365Environment
+    )
 
     Import-Module $PSScriptRoot/ProviderHelpers/CommandTracker.psm1
     $Tracker = Get-CommandTracker
@@ -105,10 +130,10 @@ function Export-AADProvider {
         # Get-PrivilegedUser provides a list of privileged users and their role assignments. Used for 2.11 and 2.12
         if ($RequiredServicePlan) {
             # If the tenant has the premium license then we want to also include PIM Eligible role assignments - otherwise we don't to avoid an API error
-            $PrivilegedUsers = $Tracker.TryCommand("Get-PrivilegedUser", @{"TenantHasPremiumLicense"=$true})
+            $PrivilegedUsers = $Tracker.TryCommand("Get-PrivilegedUser", @{"TenantHasPremiumLicense"=$true; "M365Environment"=$M365Environment})
         }
         else{
-            $PrivilegedUsers = $Tracker.TryCommand("Get-PrivilegedUser")
+            $PrivilegedUsers = $Tracker.TryCommand("Get-PrivilegedUser", @{"TenantHasPremiumLicense"=$false; "M365Environment"=$M365Environment})
         }
         $PrivilegedUsers = $PrivilegedUsers | ConvertTo-Json
         # The above Converto-Json call doesn't need to have the input wrapped in an
@@ -127,10 +152,10 @@ function Export-AADProvider {
         # Get-PrivilegedRole provides data for 2.14 - 2.16, policies that evaluate conditions related to Azure AD PIM
         if ($RequiredServicePlan){
             # If the tenant has the premium license then we want to also include PIM Eligible role assignments - otherwise we don't to avoid an API error
-            $PrivilegedRoles = $Tracker.TryCommand("Get-PrivilegedRole", @{"TenantHasPremiumLicense"=$true})
+            $PrivilegedRoles = $Tracker.TryCommand("Get-PrivilegedRole", @{"TenantHasPremiumLicense"=$true; "M365Environment"=$M365Environment})
         }
         else {
-            $PrivilegedRoles = $Tracker.TryCommand("Get-PrivilegedRole")
+            $PrivilegedRoles = $Tracker.TryCommand("Get-PrivilegedRole", @{"TenantHasPremiumLicense"=$false; "M365Environment"=$M365Environment})
         }
         $PrivilegedRoles = ConvertTo-Json -Depth 10 @($PrivilegedRoles) # Depth required to get policy rule object details
     }
@@ -248,7 +273,11 @@ function Get-PrivilegedUser {
     param (
         [ValidateNotNullOrEmpty()]
         [switch]
-        $TenantHasPremiumLicense
+        $TenantHasPremiumLicense,
+
+        [ValidateNotNullOrEmpty()]
+        [string]
+        $M365Environment
     )
 
     # A hashtable of privileged users
@@ -305,8 +334,11 @@ function Get-PrivilegedUser {
                 if ($TenantHasPremiumLicense) {
                     # Get the users that are assigned to the PIM group as Eligible members
                     # $PIMGroupMembers = Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance -All -ErrorAction Stop -Filter "groupId eq '$GroupId'"
-                    $pimGroupArgs = @{"commandlet" = "Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance"; "queryParams" = @{'$filter' = "groupId eq '$GroupId'"}}
-                    $PIMGroupMembers = Invoke-GraphDirectly @pimGroupArgs
+                    $graphArgs = @{
+                        "commandlet" = "Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance" 
+                        "queryParams" = @{'$filter' = "groupId eq '$GroupId'"}
+                        "M365Environment" = $M365Environment }
+                    $PIMGroupMembers = Invoke-GraphDirectly @graphArgs
                     foreach ($GroupMember in $PIMGroupMembers) {
                         # If the user is not a member of the PIM group (i.e. they are an owner) then skip them
                         if ($GroupMember.AccessId -ne "member") { continue }
@@ -331,7 +363,10 @@ function Get-PrivilegedUser {
     if ($TenantHasPremiumLicense) {
         # Get a list of all the users and groups that have Eligible assignments
         # $AllPIMRoleAssignments = Get-MgBetaRoleManagementDirectoryRoleEligibilityScheduleInstance -All -ErrorAction Stop
-        $AllPIMRoleAssignments = Invoke-GraphDirectly("Get-MgBetaRoleManagementDirectoryRoleEligibilityScheduleInstance")
+        $graphArgs = @{
+            "commandlet" = "Get-MgBetaRoleManagementDirectoryRoleEligibilityScheduleInstance" 
+            "M365Environment" = $M365Environment }
+        $AllPIMRoleAssignments = Invoke-GraphDirectly @graphArgs
 
         # Add to the list of privileged users based on Eligible assignments
         foreach ($Role in $AADRoles) {
@@ -385,8 +420,11 @@ function Get-PrivilegedUser {
 
                     # Get the users that are assigned to the PIM group as Eligible members
                     # $PIMGroupMembers = Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance -All -ErrorAction Stop -Filter "groupId eq '$UserObjectId'"
-                    $pimGroupArgs = @{"commandlet" = "Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance"; "queryParams" = @{'$filter' = "groupId eq '$UserObjectId'"}}
-                    $PIMGroupMembers = Invoke-GraphDirectly @pimGroupArgs
+                    $graphArgs = @{
+                        "commandlet" = "Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance"
+                        "queryParams" = @{'$filter' = "groupId eq '$UserObjectId'"}
+                        "M365Environment" = $M365Environment}
+                    $PIMGroupMembers = Invoke-GraphDirectly @graphArgs
                     foreach ($GroupMember in $PIMGroupMembers) {
                         # If the user is not a member of the PIM group (i.e. they are an owner) then skip them
                         if ($GroupMember.AccessId -ne "member") { continue }
@@ -417,17 +455,14 @@ function AddRuleSource{
         Source should be either PIM Group Name or Role Name
     #>
     param(
-        [Parameter(Mandatory=$true)]
         [ValidateNotNullOrEmpty()]
         [string]
         $Source,
 
-        [Parameter(Mandatory=$false)]
         [ValidateNotNullOrEmpty()]
         [string]
         $SourceType = "Directory Role",
 
-        [Parameter(Mandatory=$true)]
         [ValidateNotNullOrEmpty()]
         [array]
         $Rules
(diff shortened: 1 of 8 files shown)


tkol2022:
### Items Left to Complete

- [ ] Remove dependency on Microsoft.Graph.Beta.Identity.Governance from RequiredVersions.ps1 & UninstallModules.ps1

- [ ] Modify the functional tests in AAD file aad.g5.testplan.yaml to create JSON compatible with the new Rego fields for policies 7.4 & 7.5. In the Rego, the field EndDateTime was changed to endDateTime and StartDateTime was changed to startDateTime since the field that come back directly from the REST APIs start with lower case letters.
- [ ] Update the unit tests for 7.4 & 7.5 due to the renaming of the fields as described in the prior action item

- [ ] Run the tool with the current ScubaGear code from main branch and capture HTML output to compare the output against the 816 enhancement branch. Verify that the enhanced code produces the same policy output results for all section 7 policies in all the test tenants.

- [ ] Run the main branch code against AAD and then run the 816 branch and log the execution times the spreadsheet for the Tenant 2 (G3) and Tenant 6 (GCC high)
[AAD before after replacement Identity.Governance.xlsx](https://github.com/cisagov/ScubaGear/files/15278464/AAD.before.after.replacement.Identity.Governance.xlsx)

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,70 @@
...
+    "Get-MgBetaRoleManagementDirectoryRoleEligibilityScheduleInstance" = "https://graph.microsoft.com/beta/roleManagement/directory/roleEligibilityScheduleInstances"
+    "Get-MgBetaRoleManagementDirectoryRoleAssignmentScheduleInstance" = "https://graph.microsoft.com/beta/roleManagement/directory/roleAssignmentScheduleInstances"
+    "Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance" = "https://graph.microsoft.com/beta/identityGovernance/privilegedAccess/group/eligibilityScheduleInstances"
+    "Get-MgBetaPrivilegedAccessResource" = "https://graph.microsoft.com/beta/privilegedAccess/aadGroups/resources"
+    "Get-MgBetaPolicyAuthorizationPolicy" = "https://graph.microsoft.com/beta/policies/authorizationPolicy"
+    "Get-MgBetaDirectorySetting" = "https://graph.microsoft.com/beta/settings"
+    "Get-MgBetaPolicyAuthenticationMethodPolicy" = "https://graph.microsoft.com/beta/policies/authenticationMethodsPolicy"
+    "Get-MgBetaDomain" = "https://graph.microsoft.com/beta/domains"
+    "Get-MgBetaOrganization" = "https://graph.microsoft.com/beta/organization"
+    "Get-MgBetaUser" = "https://graph.microsoft.com/beta/users"
```
gdasher, line 15: We seem to have mapped some APIs (Like this one) that we are not invoking via graph directly but omitted others (such as Get-MgBetaGroup) too.  

Lets limit this to APIs that we actually invoke via graph directly.  Fine to do a test of more if you want to but we should have a method.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,70 @@
...
+    "Get-MgBetaPolicyRoleManagementPolicyAssignment"  = "https://graph.microsoft.com/beta/policies/roleManagementPolicyAssignments"
+}
+
+function Invoke-GraphDirectly {
```
gdasher, line 19: Can you please add a test case for this function that mocks out Invoke-MgGraphRequest?

tkol2022: This comment refers to a Powershell unit test. Touch base with @schrolla with specifics on how to do that.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,70 @@
...
+        [ValidateNotNullOrEmpty()]
+        [string]
+        $M365Environment,
+
+        [System.Collections.Hashtable]
+        $queryParams
+    )
+
+    Write-Debug "Replacing Cmdlet: $commandlet"
+    $endpoint = $GraphEndpoints[$commandlet]
```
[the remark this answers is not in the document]
twneale: Added this

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,70 @@
...
+    # GCC high and DoD have slightly different endpoint URLs
+    if (($M365Environment -eq "gcchigh") -or ($M365Environment -eq "dod")) {
```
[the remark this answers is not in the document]
twneale: Did a quick test--and Invoke-MgGraphRequest *does* add the rest of the URL if only a path is provided. I tried removing the domain part of the URL and use paths instead. It works! Just need to test it on gcchigh and should be all set.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -56,10 +130,10 @@ function Export-AADProvider {
         # Get-PrivilegedUser provides a list of privileged users and their role assignments. Used for 2.11 and 2.12
         if ($RequiredServicePlan) {
             # If the tenant has the premium license then we want to also include PIM Eligible role assignments - otherwise we don't to avoid an API error
-            $PrivilegedUsers = $Tracker.TryCommand("Get-PrivilegedUser", @{"TenantHasPremiumLicense"=$true})
+            $PrivilegedUsers = $Tracker.TryCommand("Get-PrivilegedUser", @{"TenantHasPremiumLicense"=$true; "M365Environment"=$M365Environment})
```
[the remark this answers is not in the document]
twneale: After converting the URLs to paths only, I was able to delete these and it still worked. Will test on gcchigh to be sure.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,70 @@
+$GraphEndpoints = @{
```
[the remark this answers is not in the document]
twneale: In the tests I did, all of the commandlets in the GraphEndpoints object were replaced with direct API calls. Many of the API calls are faster than the commandlets, but a few offered no performance improvement over the commandlets. Ted thought it was better to start moving them over with the slowest first and confirm nothing breaks.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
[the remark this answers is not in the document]
twneale: It worked!

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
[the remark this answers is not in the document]
twneale: Gcchigh test worked

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -332,7 +419,12 @@ function Get-PrivilegedUser {
                     }
 
                     # Get the users that are assigned to the PIM group as Eligible members
-                    $PIMGroupMembers = Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance -All -ErrorAction Stop -Filter "groupId eq '$UserObjectId'"
+                    # $PIMGroupMembers = Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance -All -ErrorAction Stop -Filter "groupId eq '$UserObjectId'"
```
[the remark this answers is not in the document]
twneale: I added a comment at the top where we list the mapping of commandlets to API endpoints that replace them.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
twneale: I whittled it down to the four that we are currently using. The rest can live on in the git history until we need them

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
twneale: Added a basic test to cover er this function

tkol2022:
### For reviewers: How to test the performance improvement
The performance improvement when running AAD is mostly noticeable when the tool executes for the first time in a newly spawned Powershell window. This is because the slowness that we observed in the Microsoft.Graph.Beta.Identity.Governance dll was during its load into memory, which occurs only the first time you call one of its Cmdlets in a new Powershell window.
The steps below describe how you can test to ensure that you also see an improvement.

Use this code to take a time measurement (tailor the parameter values to your environment):
```
Import-Module -Name .\PowerShell\ScubaGear
$ExecutionTime = Measure-Command { Invoke-Scuba -ProductNames aad -CertificateThumbPrint "33dd41a761f434f12ee8ac3ff771347597435585" -AppID "ac691fde-1ff0-493f-be3d-03d9ba6cb891" -Organization "TEST_TENANT.onmicrosoft.com" }
$ExecutionTime
```

1. Download a copy of the main branch and run the measurement code above 5-10 times. Remember to take each measurement in a new Powershell window. Compute an average time.
2. Download a copy of the branch associated with this PR and repeat step 1.
3. Log the average time of the main branch and this branch in this PR.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,51 @@
...
+        <#
+          If query params are passed in, we need to augment the endpoint URI to include the params. Paperwork below.
+          We can't use string formatting easily because the graph API expects parameters
+          that are prefixed with the dollar symbol. Maybe someone smarter than me can figure that out.
+        #>
+        $q = [System.Web.HttpUtility]::ParseQueryString([string]::Empty)
+        foreach ($item in $queryParams.GetEnumerator()) {
+            $q.Add($item.Key, $item.Value)
+        }
+        $endpoint = "https://example.com" + $endpoint
```
gdasher, line 38: ???

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -1,3 +1,51 @@
...
+        $uri = [System.UriBuilder]$endpoint
+        $uri.Query = $q.ToString()
+        $endpoint = $uri.Path + $uri.Query
```
gdasher, line 41: This seems off.  

You should be able to build the Uri using host, query string, scheme, etc and then use uri.ToString():

https://learn.microsoft.com/en-us/dotnet/api/system.uribuilder?view=net-8.0

PowerShell/ScubaGear/Testing/Unit/PowerShell/Providers/AADProvider/Invoke-GraphDirectly.ps1
```diff
@@ -0,0 +1,22 @@
+$ProviderPath = '../../../../../Modules/Providers'
+Import-Module (Join-Path -Path $PSScriptRoot -ChildPath "$($ProviderPath)/ExportAADProvider.psm1") -Function 'Get-PrivilegedRole' -Force
+
+InModuleScope ExportAADProvider {
+    Describe -Tag 'AADProvider' -Name "Invoke-GraphDirectly" {
+        BeforeAll {
+            #function Invoke-MgGraphRequest {return @{Value = @{cowsound = "moo"}}
```
gdasher, line 7: <3

gdasher (changes requested):
Thanks for the changes.  One remaining area around buildilng the Uri and dispatching to the different environments.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
adhilto: I was also concerned when I saw this. Digging in deeper, since the endpoint ends up being the path and query concatenated together as a string, the host value itself doesn't matter. But surely there's a way to construct the endpoint string without specifying an arbitrary third-party host, even if just temporarily?

adhilto: If nothing else, you should add a comment stating that the value for host isn't actually used, i.e., no data will actually be sent to example.com. Additionally, you could specify localhost as the host, that might be at least slightly less concerning to any users reading this code.

adhilto (commented):
My two cents: this is faster. I ran AAD once on this branch and once on main: 108 seconds on this branch and 178 on main, so over a minute faster. And manual inspection of the two reports showed no difference in the output, as expected.

One minor piece of feedback beyond what has already been commented on: a lot of the new code does not use PascalCase, per our style guide. Not a big deal but would be nice to have for consistency.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
twneale: @adhilto Good point. I tried leave the scheme and host as empty strings and it gives the exact result we want. I will change it.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
twneale: I will change it.

twneale: It works passing all 5 params interactively in powershell, but refuses the 5th parameter (query string) when I add this to the Scuba code. I'm at a loss for this goofy behavior, so I settled on passing all of them in and concat'ing the query string. I will push the changes so everyone can see them.

PowerShell/ScubaGear/Testing/Unit/PowerShell/Providers/AADProvider/Invoke-GraphDirectly.ps1
twneale: Should I change it to a different barnyard denizen?

twneale:
Here are the run durations I recorded:

[durationtimes.xlsx](https://github.com/user-attachments/files/15742518/durationtimes.xlsx)

Of special note: many times when I ran the tests on either branch, `invoke-scuba` impossibly exited fine within 1-5 seconds. I need to disclose that I thoroughly perjured the scientific integrity of these test by completely ignoring those runs. I don't know how they went so fast.

The conclusion: for some reason, running in Tenant 2 conferred a minor performance advantage. In gcchigh it was about 60% faster on average.

PowerShell/ScubaGear/Modules/Providers/ExportAADProvider.psm1
```diff
@@ -332,7 +388,12 @@ function Get-PrivilegedUser {
                     }
 
                     # Get the users that are assigned to the PIM group as Eligible members
-                    $PIMGroupMembers = Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance -All -ErrorAction Stop -Filter "groupId eq '$UserObjectId'"
+                    # $PIMGroupMembers = Get-MgBetaIdentityGovernancePrivilegedAccessGroupEligibilityScheduleInstance -All -ErrorAction Stop -Filter "groupId eq '$UserObjectId'"
```
adhilto, line 391: Is there a good reason to keep this commented out line? Otherwise recommend for deletion.
```suggestion
```

twneale: I thought it might be good to keep these so people would understand what the `Invoke-GraphDirectly` function is doing, and easily revert them if issues arise but...that's probably not a good reason. I will nuke these.

tkol2022:
> Here are the run durations I recorded:
> 
> [durationtimes.xlsx](https://github.com/user-attachments/files/15742518/durationtimes.xlsx)
> 
> Of special note: many times when I ran the tests on either branch, `invoke-scuba` impossibly exited fine within 1-5 seconds. I need to disclose that I thoroughly perjured the scientific integrity of these test by completely ignoring those runs. I don't know how they went so fast.
> 
> The conclusion: for some reason, running in Tenant 2 conferred a minor performance advantage. In gcchigh it was about 60% faster on average.

Make sure that when you run the timing, each timing has to be done in a fresh PS window because it is the loading of the dependency DLL which accounts for the time we think we are saving with our new code. The DLL loading only occurs when you first Invoke-Scuba the first time in a new window.

adhilto (approved):

tkol2022:
@gdasher @mitchelbaker-cisa Can you quickly review if there is anything left to change on this PR? If not, we will push it to get merged.

mitchelbaker-cisa (approved):
Ran tests against G5 tenant and saw almost a minute improvement. 

Average execution time over 9 tests:

Current AAD: 2 min 28 sec
AAD Identity.Governance replacement: 1 min 34 sec

tkol2022:
I'm closing this and opened a fresh one #1196. Handling the merge conflicts with the AAD unit test refactor proved a losing battle.

Issue #740, opened 2023-12-13
Report on ignored policies


crutchfield:
# 💡 Summary #

As a user, I want to be able to direct ScubaGear to ignore evaluating and reporting on compliance with a given policy. I want ScubaGear to report the policy is not being evaluated for compliance and show my provided rationale.

## Motivation and context ##

The user in a realistic operational environment may choose to accept risk of non-compliance with a given policy. Our tool should support the user as they manage these risks.

This would be useful because it will allow the user to get a report of a document choice to ignore compliance with a given policy.

## Implementation notes ##

Details for implementation:

- An ignored policy is reported as such with a provided rationale

## Acceptance criteria ##

How do we know when this work is done?

- [ ] Any policy specified to be ignored for a given reason is reported as such

tkol2022:
Not sure why but the description for #739 seems identical with this one except that 739 has some additional scope. If this is true, then close this issue as a duplicate.

Issue #739, opened 2023-12-13
Allow any given policy to be ignored


crutchfield:
# 💡 Summary #

As a user, I wnat to be able to direct ScubaGear to ignore evaluating and reporting on compliance with a given policy. I want ScubaGear to report the policy is not being evaluated for complaince and show my provided rationale. 

## Motivation and context ##

The user in a realistic operational environment may choose to accept risk of non-compliance with a given policy.  Our tool should support the user as they manage these risks. 

This would be useful because it will allow the user to document explicit choices to ignore complaince with a given policy.

## Implementation notes ##

Details for implementation:

- Configuration file allows for ignoring a policy
- Command line allow for ignoring a policy
- Configuration file allows for providing rationale for ignoring a policy
- Command line allows for providing rationale for ignoring a policy
- A policy specified to be ignored is not evaluated 
- An ignored policy is reported as such with a provided rationale 

## Acceptance criteria ##

- [ ] Any policy specified to be ignored far a given reason is ignored
- [ ] Documentation on use of ignore capability is provided

Issue #738, opened 2023-12-13
Policy item toggling feature


crutchfield:
# 💡 Summary #

The user should be able to specify a set of policies not to be evaluated by ScubaGear.  

## Motivation and context ##

Normal Ops can identify policy deviations that are of acceptable risk for their organization.   

This would be useful because the user can manage which policies they are monitoring.  

## Implementation notes ##

Please provide details for implementation, such as:

- User can provide configuration item to ignore a policy for evaluation
- User can provide a rationale for ignoring a policy for evaluation
- Report should shows that a policy is ignored and why (rationale) 

## Acceptance criteria ##

The following issues are completed

- [ ] Allow any given policy to be ignored #739 
- [ ] Report ignored policies with rationale #740

PR #1200, opened 2024-07-05
Allow policies to be omitted from ScubaGear

adhilto:
## 🗣 Description ##
Make it so that policies can be omitted from the ScubaGear reports.

Closes https://github.com/cisagov/ScubaGear/issues/740.
Closes https://github.com/cisagov/ScubaGear/issues/739.
Closes https://github.com/cisagov/ScubaGear/issues/738.

## 🧪 Testing ##
I'd recommend that the reviewers start by reading the documentation I added for this feature, to get a good overview of the approach taken: https://github.com/cisagov/ScubaGear/blob/739-allow-any-given-policy-to-be-ignored/docs/configuration/configuration.md#omit-policies.

Testing I've done:
- Added 8 new Pester cases to cover these changes
- Manually tested without a config file to ensure normal executions is unchanged
- Manually tested excluding several policies without expiration dates
- Manually tested excluding policies without specifying the rationale
- Manually tested excluding policies with expirations date, both in the past and future
- Manually tested excluding policies with a malformed expiration date
- Manually tested generating a config file with policy omissions via New-Config, including some edge cases, such as malformed policy ids, variations in capitalization, and unexpected product names in the policy IDs.

## ✅ Pre-merge checklist ##

- [ ] PR passed smoke test check.
- [ ] Feature branch has been rebased against changes from parent branch, as needed

- [ ] Resolved all merge conflicts on branch

## ✅ Post-merge checklist ##

### DIFF
--- a/PowerShell/ScubaGear/Modules/Orchestrator.psm1
+++ b/PowerShell/ScubaGear/Modules/Orchestrator.psm1
@@ -1005,8 +1005,12 @@ function Invoke-ReportCreation {
                 $LinkPath = "$($IndividualReportFolderName)/$($BaselineName)Report.html"
                 $LinkClassName = '"individual_reports"' # uses no escape characters
                 $Link = "<a class=$($LinkClassName) href='$($LinkPath)'>$($FullName)</a>"
-
-                $PassesSummary = "<div class='summary pass'>$($Report.Passes) tests passed</div>"
+                if ($Report.Skips -gt 0) {
+                    $PassesSummary = "<div class='summary pass'>$($Report.Passes) passed / $($Report.Passes) skipped</div>"
+                }
+                else {
+                    $PassesSummary = "<div class='summary pass'>$($Report.Passes) tests passed</div>"
+                }
                 $WarningsSummary = "<div class='summary'></div>"
                 $FailuresSummary = "<div class='summary'></div>"
                 $BaselineURL = "<a href= `"https://github.com/cisagov/ScubaGear/blob/v$($ModuleVersion)/baselines`" target=`"_blank`"><h3 style=`"width: 100px;`">Baseline Documents</h3></a>"
--- a/PowerShell/ScubaGear/Modules/Orchestrator.psm1
+++ b/PowerShell/ScubaGear/Modules/Orchestrator.psm1
@@ -1005,17 +1005,18 @@ function Invoke-ReportCreation {
                 $LinkPath = "$($IndividualReportFolderName)/$($BaselineName)Report.html"
                 $LinkClassName = '"individual_reports"' # uses no escape characters
                 $Link = "<a class=$($LinkClassName) href='$($LinkPath)'>$($FullName)</a>"
-                if ($Report.Skips -gt 0) {
-                    $PassesSummary = "<div class='summary pass'>$($Report.Passes) passed / $($Report.Skips) skipped</div>"
-                }
-                else {
-                    $PassesSummary = "<div class='summary pass'>$($Report.Passes) tests passed</div>"
-                }
+                $PassesSummary = "<div class='summary'></div>"
                 $WarningsSummary = "<div class='summary'></div>"
                 $FailuresSummary = "<div class='summary'></div>"
                 $BaselineURL = "<a href= `"https://github.com/cisagov/ScubaGear/blob/v$($ModuleVersion)/baselines`" target=`"_blank`"><h3 style=`"width: 100px;`">Baseline Documents</h3></a>"
                 $ManualSummary = "<div class='summary'></div>"
-                $ErrorSummary = "<div class='summary'></div>"
+                $OmitSummary = "<div class='summary'></div>"
+                $ErrorSummary = ""
+
+                if ($Report.Passes -gt 0) {
+                    $Noun = Pluralize -SingularNoun "pass" -PluralNoun "passes" -Count $Report.Passes
+                    $PassesSummary = "<div class='summary pass'>$($Report.Passes) $($Noun)</div>"
+                }
 
                 if ($Report.Warnings -gt 0) {
                     $Noun = Pluralize -SingularNoun "warning" -PluralNoun "warnings" -Count $Report.Warnings
@@ -1023,13 +1024,17 @@ function Invoke-ReportCreation {
                 }
 
                 if ($Report.Failures -gt 0) {
-                    $Noun = Pluralize -SingularNoun "test" -PluralNoun "tests" -Count $Report.Failures
-                    $FailuresSummary = "<div class='summary failure'>$($Report.Failures) $($Noun) failed</div>"
+                    $Noun = Pluralize -SingularNoun "failure" -PluralNoun "failures" -Count $Report.Failures
+                    $FailuresSummary = "<div class='summary failure'>$($Report.Failures) $($Noun)</div>"
                 }
 
                 if ($Report.Manual -gt 0) {
                     $Noun = Pluralize -SingularNoun "check" -PluralNoun "checks" -Count $Report.Manual
-                    $ManualSummary = "<div class='summary manual'>$($Report.Manual) manual $($Noun) needed</div>"
+                    $ManualSummary = "<div class='summary manual'>$($Report.Manual) manual $($Noun)</div>"
+                }
+
+                if ($Report.Omits -gt 0) {
+                    $OmitSummary = "<div class='summary manual'>$($Report.Omits) omitted</div>"
                 }
 
                 if ($Report.Errors -gt 0) {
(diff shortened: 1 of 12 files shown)


PowerShell/ScubaGear/Modules/Orchestrator.psm1
gdasher, line 1: It looks like as implemented this won't change the Json version of the report.  I think that omissions should be included in the json output as well, that way for example we can build graphs to track them historically in MTA without rejoining with the configuration state (which may have changed from the moment of assessment to present).

adhilto: Great callout. It does* (I just double checked to be safe).

The asterisk: it changes the ScubaResults.json file, but not the Rego output, TestResults.json, which is probably what you've picked up on.

ScubaResults.json is the json equivalent of the HTML report and is the file used for MTA. Here's an example of what it looks like when run with an omission:
image

TestResults.json is just an intermediary file (arguably with a misleading name) and isn't used for MTA.

gdasher: Ahh I see now how the new version works.  Ok, thanks for checking.  Looks good!

tkol2022:
### Feedback on example config file

The example MS.TEAMS.6.1v1 provided in the sample config file is confusing with respect to the sample rationale. If the service is provided by M365 Defender and the Defender baseline checks for the implementation of this policy, then what is the purpose of having this policy in Teams to begin with? I believe this has a good chance of confusing users.
Instead I recommend you provide an example rationale such as "The DLP capability required for Teams is implemented by third party product X which ScubaGear does not have the ability to check" or something like that.

I also recommend that you trim the number of example policy IDs that you have. I think by the time the user gets to policy 7.1 in the example config they will get the point so you could get rid of examples 7.2, 8.1, 8.2 and we wouldn't lose anything.

adhilto:
> ### Feedback on example config file
> The example MS.TEAMS.6.1v1 provided in the sample config file is confusing with respect to the sample rationale. If the service is provided by M365 Defender and the Defender baseline checks for the implementation of this policy, then what is the purpose of having this policy in Teams to begin with? I believe this has a good chance of confusing users. Instead I recommend you provide an example rationale such as "The DLP capability required for Teams is implemented by third party product X which ScubaGear does not have the ability to check" or something like that.
> 
> I also recommend that you trim the number of example policy IDs that you have. I think by the time the user gets to policy 7.1 in the example config they will get the point so you could get rid of examples 7.2, 8.1, 8.2 and we wouldn't lose anything.

Done

tkol2022:
I made a few tweaks to the configuration.md because i felt the term "excluded" could get conflated with AAD conditional access policy exclusions. we should stick with the term "omitted" and omissions.

tkol2022:
In configuration.md, I think we can delete this paragraph. It is confusing and I don't think it adds anything. Do you think anyone is likely to send policy ids as command line arguments to New-Config? It is easier to just modify the yaml config file in my opinion.

"Policy omissions can be provided to the `New-Config` function as a comma-separated list of policy IDs under the `OmitPolicy` parameter. However, the `New-Config` function does not allow you to specify the rationales or expiration dates via the commandline; these must be manually entered into the resulting config file."

tkol2022:
The expiration date feature is a nice touch!

tkol2022:
Will the following line perform a case-insensitive match? If not, should we make is so that case does not matter?

`if ($Config.OmitPolicy.psobject.properties.name -Contains $ControlId) `

tkol2022:
What are your thoughts on changing the following?
Test disabled by user.
to
Test omitted by user.

tkol2022:
### Bug
If the user accidentally provides two Rationale fields ScubaGear crashes. The same problem occurs if you have more than one omit policy entry for the same policy id.

image
image

image

tkol2022:
> Will the following line perform a case-insensitive match? If not, should we make is so that case does not matter?
> 
> `if ($Config.OmitPolicy.psobject.properties.name -Contains $ControlId) `

I tested this. It ignores case which is great for flexibility. No change needed.
image

tkol2022:
### Test summary

- Tested prior date, today and future date for the Expiration field and the tool behaved as expected.
- Tested single quote in Rationale field and tool behaved as expected.
- Tested rationale missing and tool behaved as expected.
- Tested malformed date and the tool behaved as expected.
- Tested without rationale and without date and the tool behaved as expected.
- The only test that produced a bug was when you have more than one of the same policy or you have multiple rationales. See separate comment with details.

tkol2022 (approved):
I found one bug and left some comments for a couple of other tweaks. If you can address those I think we are in good shape. Nice work. Overall the system behaved as expected under various conditions. Also I asked a related question on Slack which was answered by the team about how the downstream system interprets the outputs from this new functionality.

PowerShell/ScubaGear/Modules/Support/Support.psm1
```diff
@@ -849,6 +863,46 @@ function New-Config {
...
 
+    $OmissionNamespace = "OmitPolicy"
+
+    # List to track which policies the user specified in $OmitPolicies are properly formatted
+    $OmitPolicyValidated = @()
+
+    # Hashmap to structure the ignored policies template
+    $config[$OmissionNamespace] = @{}
+
+    foreach ($Policy in $OmitPolicy) {
```
mitchelbaker-cisa, line 874: This logic validates the formatting of $OmitPolicies when a user first invokes New-Config, but I'm not seeing any validation in cases where the config.yml is modified after creation. If I incorrectly set a key in `OmitPolicy` to `MSAAD4.1v1` or `MS.GWS.AAD4.2v1` these omissions are skipped and the policy runs normally. 

Perhaps an additional unit test to verify malformed policy IDs are excluded from CreateReport.psm1 and the policy is expected to run by default?

mitchelbaker-cisa (commented):
Testing:

Ran ScubaGear without a config file, runs as expected. 
Ran ScubaGear with a config file:
- with valid policy ID, rationale, expiration
- with invalid policy ID
- with null Expiration value (results in skipped omission for given policy)
- with non-string Expiration value (results in skipped omission for given policy)

Verified "OmitPolicy" is included in ProviderSettingsExport 
Verified Result: "Omitted" is in individual report json and in combined json with -mergeJson flag

Left one comment to consider but otherwise looks good.

adhilto:
> In configuration.md, I think we can delete this paragraph. It is confusing and I don't think it adds anything. Do you think anyone is likely to send policy ids as command line arguments to New-Config? It is easier to just modify the yaml config file in my opinion.
> 
> "Policy omissions can be provided to the `New-Config` function as a comma-separated list of policy IDs under the `OmitPolicy` parameter. However, the `New-Config` function does not allow you to specify the rationales or expiration dates via the commandline; these must be manually entered into the resulting config file."

I agree that it's probably easier to reference the example config and modify their config file manually. I'm ok with deleting the paragraph.

adhilto:
> ### Bug
> If the user accidentally provides two Rationale fields ScubaGear crashes. The same problem occurs if you have more than one omit policy entry for the same policy id.

I think this bug is out of scope for this PR, I recommend opening a separate issue for it. This bug is in the orchestrator and/or the ScubaConfig.psm1 `LoadConfig` function. The YAML specification actually does not allow duplicate keys, so in your test, you've engineered invalid yaml. The problem this surfaces is that the orchestrator does not catch that invalid yaml was provided. The first error message in your screenshot is a bit cryptic, but the fix might be to wrap this line (in ScubaConfig.psm1) in a try/catch:
image

Additionally, for whatever reason, this `if` in Orchestraror.psm1 failed to detect that the config failed to load.
image

tkol2022:
> Additionally, for whatever reason, this `if` in Orchestraror.psm1 failed to detect that the config failed to load. image

I opened an issue. BTW, from what I can tell, the reason that code in Orchestrator does not work as expected is because when the LoadConfig function produces an error, Orchestrator will not execute the if statement you mentioned. Instead it will go to the nearest catch () block, right?

adhilto:
> > Additionally, for whatever reason, this `if` in Orchestraror.psm1 failed to detect that the config failed to load. image
> 
> I opened an issue. BTW, from what I can tell, the reason that code in Orchestrator does not work as expected is because when the LoadConfig function produces an error, Orchestrator will not execute the if statement you mentioned. Instead it will go to the nearest catch () block, right?

Honestly, I'm not really sure what will happen. Unless I'm just not looking deep enough, this block of code isn't within a `try` at all. Definitely more digging needed here.

PowerShell/ScubaGear/Modules/Support/Support.psm1
adhilto: Done. I also modified the load config function to output warnings for invalid IDs. Tagging @tkol2022 for awareness.
image

adhilto:
> What are your thoughts on changing the following? Test disabled by user. to Test omitted by user.

Changed

mitchelbaker-cisa (approved):

adhilto:
@nanda-katikaneni Looks like this one is good to go

PR #1406, opened 2024-11-06
improve performance of Defender query to count users without advanced auditing

### LINKED ISSUE #1404: Defender export provider inadvertently downloads user data when retrieving a count of users without advanced auditing
# 💡 Summary #

Recently we noticed that when the tenant size becomes larger (e.g. 20,000 users) the Defender provider takes significantly longer to execute than previously observed and that prompted my code review which resulted in this issue.

The Defender export provider contains code with calls the server-side API associated with Get-MgBetaUser using a filter to get a count of users that do not have the advanced auditing feature enabled. The way the code is structured, the only piece of data that ends up being used is a result count value which is passed to the Rego policy checks. I manually ran the cmdlet with the same parameters used in the code against a test tenant and noticed that it not only provides data in the count variable, but the cmdlet also returns a dataset of all the users that match the filter with a comprehensive set of properties in Powershell objects. This leads me to believe that for large tenants, particularly ones that have a lot of users that are NOT assigned the advanced audit feature, this code may take some time to complete because it is downloading the dataset from the server. It could also be the case that downloading the data is fast and the query simply takes a long time to complete on the server side so that should be investigated.

I think this should be investigated and anything that we can do to improve the performance of this policy check for larger tenants should be considered. This is especially pertinent because many large tenants might not have the advanced audit feature turned on for a wide subset of users.

Here is the code that performs the count against the server using a Filter parameter.
Image

Observe below when I ran the same command manually against a test tenant that it returns the full dataset of matching users and their properties. Even though we are not using this information in the code because the cmdlet is being piped to Out-Null, the data is still being downloaded which could take time, not to mention a Powershell object is being created for every downloaded user with a host of properties in RAM.

`Get-MgBetaUser -Filter "not assignedPlans/any(a:a/servicePlanId eq 2f442157-a11c-46b9-ae5b-6e39ff4e5849 and a/capabilityStatus eq 'Enabled')" -Count UsersWithoutAdvancedAuditCount -ConsistencyLevel eventual -All`

Image

Each downloaded user gets its own Powershell object each containing 128 property keys and associated values:
Image

## Motivation and context ##

Improving the performance of the tool is paramount to user satisfaction. Also with larger tenants this code seems to significantly slow down our automated functional tests when running Defender so if we correct this it will enable us to test features for larger tenants without performance impacts.

tkol2022:
## 🗣 Description ##

This PR fixes a bug in Defender that is observed in larger tenants. Defender runs a query via Get-MgBetaUser to get a count of users that do not have the advanced auditing feature assigned to them. Instead of just retrieving the count from the REST API, the current query downloads all the users that match the filter. If a tenant has thousands of users without advanced audit this can result in significant slowness. I added the Top = 1 parameter so that only a single record is returned and verified that the count still works. See the linked issue below for full details and screenshots.

Closes #1404 

## 🧪 Testing ##

See the screenshots in the linked issue above for details on the before and after testing that I performed with the fix against a tenant with over 2,000 users.

## ✅ Pre-merge checklist ##

## ✅ Post-merge checklist ##

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportDefenderProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportDefenderProvider.psm1
@@ -143,13 +143,13 @@ function Export-DefenderProvider {
         $DLPLicense = ConvertTo-Json $false
     }
 
-    # Run commands to get count of users with Advanced auditing enabled
+    # Get count of users without Advanced auditing enabled
     # GUID below is service plan ID for M365_ADVANCED_AUDITING as defined
     # on Microsoft Licensing Reference shown here:
     # https://learn.microsoft.com/en-us/entra/identity/users/licensing-service-plan-reference
     $UserParameters = @{ConsistencyLevel = 'eventual'
                         Count = 'UsersWithoutAdvancedAuditCount'
-                        All = $true
+                        Top = 1
                         Filter = "not assignedPlans/any(a:a/servicePlanId eq 2f442157-a11c-46b9-ae5b-6e39ff4e5849 and a/capabilityStatus eq 'Enabled')"
                        }
 


tkol2022:
## Testing Instructions for Reviewers
The goals of testing this PR are to ensure:

- The new code produces an accurate count of users without advanced audit assigned.
- The new code runs faster than the old code.

- Test tenant 3 has been prepared with over 2,000 users to help create a "larger" tenant scenario and you should definitely test against it, but it would be good to test the code against some of the other tenants as well to ensure the count gets calculated as expected and tease out any nuances that would only be seen in other tenant types.

### Install dependencies
Install the Powershell profiler module.
`Install-Module Profiler`

### Take measurements using the main branch (status quo)

1. Download the main branch
2. Run the profiler against the main branch when running Defender
```
Import-Module -Name .\PowerShell\ScubaGear

$trace = trace-script -scriptblock { Invoke-SCuBA -ProductNames defender -CertificateThumbPrint "55fd41a761f434f12ee8ac3ff771347597435641" -AppID "ab791fde-1ff0-493f-be3d-03d9ba6cb091" -Organization "contoso.onmicrosoft.com" }

$trace.top50duration | fl | Out-File -FilePath .\MainBranchPerformance.txt
```

3. Open the file MainBranchPerformance.txt and search for the profiler data for the Get-MgBetaUser cmdlet and jot down the value from the Duration field:

image

4. Open the output file ScubaResults.json and search for the field total_users_without_advanced_audit. Jot down the value of the field.

image

### Take measurements using the fix branch (associated with this PR)
1. Download the fix branch
2. Repeat the same steps that you performed against the main branch and take measurements using the fix branch code.

### Validate that the fix works as expected
1. Compare the execution times of the Get-MgBetaUser cmdlet between the main branch versus the fix branch. Ensure that the fix branch is faster. Note: a more pronounced difference in execution time will be observed when the tenant is larger.
2. Compare the counts of the total_users_without_advanced_audit field in ScubaResults.json between the main branch versus the fix branch. They should be identical.

james-garriss (approved):
Tested profiler in 2 tenants and found that code improved runtime.

tkol2022:
@nanda-katikaneni ready for merge

Issue #1422, opened 2024-11-12
Append UUID to ScubaResults.json by default


buidav:
# 💡 Summary #
Request to append the ReportUUID to the end of the `ScubaResults.json` by default. 

## Motivation and context ##
Request from internal teams.

## Implementation notes ##
- Requires changing the output file name in the MergeJSON function. 

## Acceptance criteria ##
- `ScubaResults.json` => ScubaResults_{UUID}.json
- Tests continue to pass even with the file name change.

Issue #1354, opened 2024-10-04
Ensure that file name parameters don't include paths


adhilto:
# 💡 Summary #
ScubaGear should ensure that the file name parameters that aren't intended to include a path don't include a path, to ensure that all ScubaGear output is saved inside the specified output folder.

## Motivation and context ##
Want to avoid a user inadvertently overwriting a file outside of the output folder.

## Implementation notes ##
The following parameters fall in this category:
- OutJsonFileName
- OutCsvFileName
- OutPlanFileName
- OutRegoFileName
- OutProviderFileName

## Acceptance criteria ##
- [ ] A check for those parameters has been added

Issue #1254, opened 2024-08-01
Add a sample config config file that includes all possible parameters including per product namespace parameters


buidav:
# 💡 Summary #

When looking at the [full_config.yaml](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/Sample-Config-Files/full_config.yaml) sample config file. 

I expected to see a config file containing all possible config files values including per product namespace variables (aad, defender), omitting polices, etc. 

However, it only contains all of our general parameters. 
The "full" name now is misleading. 

## Motivation and context ##

The config file is required to meet policy checks in AAD and Defender. 
We should have an example config file containing every possible value. 
For ease of viewing and understanding. 
Without having a user have to pull up multiple sample config files to understand the options available to them. 

## Implementation notes ##
Create a sample SCuBA Config file to stick in our sample folder and point to in our config file [documentation](https://github.com/cisagov/ScubaGear/blob/main/docs/configuration/configuration.md).

Stick the sample config file in the current sample config file folder located here
`PowerShell/ScubaGear/Sample-Config-Files`

Our `New-SCuBAConfig` cmdlet can be used to generate this file.
It is located in `PowerShell/ScubaGear/Modules/Support/Support.psm1`
Update `New-SCuBAConfig` for any missing parameters and clean up any artifacts there

I'm also noticing dead code located in that function. 
<img width="455" alt="deadCode" src="https://github.com/user-attachments/assets/01e901e6-fae3-4a5b-b53b-ae89aee90d6e">

## Acceptance criteria ##
- When an example of a full SCuBA config is in the sample config folder with documentation updated to point to it. 
- `New-SCuBAConfig` is updated with dead code removed.

DickTracyII:
There are two functions within the UI that has parameter called, however PSScript Analyzer does not detect them. I added these to suppress the warning. 

- Function: `New-FieldListCard` have this issue for `ProductName` and `OutputData`

  These parameters ARE used here:
  
  ```powershell
  # Initialize policy level if not exists
  if (-not $OutputData[$ProductName]) {
      $OutputData[$ProductName] = @{}
  }
  if (-not $OutputData[$ProductName][$policyId]) {
      $OutputData[$ProductName][$policyId] = @{}
  }
  ```
  and later down here:

  ```powershell
  # If the inputType has no value, store fields directly under the policy
  foreach ($fieldKey in $fieldCardData.Keys) {
[...]

PR #1426, opened 2024-11-14
Add the report UUID to the ScubaResults.json filename

adhilto:
## 🗣 Description ##
- Add the report UUID to the ScubaResults.json filename.
- Adds a new parameter `-NumberOfUUIDCharactersToTruncate` to allow for trunacation or omission of the UUID appneded to the file name.

TODOs discovered over the course of this issue.
#1254
#1354 
#1437
#1438 

Ensures each results file has a unique file name. Closes https://github.com/cisagov/ScubaGear/issues/1422.

## 🧪 Testing ##
- Ensured PowerShell unit tests pass
- Added new unit tests to cover the case when there are multiple matching ScubaResults*.json files
- Manually tested `Invoke-Scuba`, works as expected
- Manually tested `Invoke-ScubaCached`, works as expected
- Manually tested `Invoke-ScubaCached` where the input folder has one file matching "ScubaResults*.json", "ScubaResults.json", works as expected
- Manually tested `Invoke-ScubaCached` where the input folder has one file matching "ScubaResults*.json", "ScubaResults_{uuid}.json", works as expected
- Manually tested `Invoke-ScubaCached` where the input folder has two file matching "ScubaResults*.json", the most recently created file is used, as expected
- Manually tested `Invoke-Scuba` and `Invoke-ScubaCached` with a non-default `OutPath` parameter
- Manually tested both `Invoke-Scuba` and `Invoke-ScubaCached` with a non-default `OutJsonFileName` parameter

## ✅ Pre-merge checklist ##

- [x] Notified merge coordinator that PR is ready for merge via comment mention
- [x] Demonstrate changes to the team for questions and comments. 
    (Note: Only required for issues of size `Medium` or larger)

## ✅ Post-merge checklist ##


tkol2022:
Should we make the NumberOfUUIDCharactersToTruncate and KeepIndividualJson parameters mutually exclusive? I was thinking if you pass KeepIndividualJson then it doesn't make sense to consider the NumberOfUUIDCharactersToTruncate? Should the tool provide an error alerting the user?

If we decide to implement this, we can add this as an enhancement later because I don't feel comfortable making too many changes given the importance of this release.

image

tkol2022:
We might want to do some data validation on the OutJsonFileName parameter. If you pass "*" as the name, it crashes.
Maybe add a ValidatePattern to the parameter definition in Invoke-Scuba?

```
[Parameter(Mandatory = $false, ParameterSetName = 'Configuration')]
[Parameter(Mandatory = $false, ParameterSetName = 'Report')]
[ValidateNotNullOrEmpty()]
[ValidatePattern('^[a-zA-Z0-9]+$')]
[string]
$OutJsonFileName = [ScubaConfig]::ScubaDefault('DefaultOutJsonFileName')
```

Again this is an enhancement I think we can make in the future but wanted to discuss.

image

tkol2022:
I see that we are truncating the filename GUID in multiple places, ConvertTo-ResultsCsv and Merge-JsonOutput resulting in duplicate logic that we have to keep in synch over time. What if we truncate the GUID in a helper function? Is it worth it?

image

image

Also, the Merge-JsonOutput function gets the GUID from the line of code below which gets it from the settings export object, whereas the ConvertTo-ResultsCsv function gets the GUID from a parameter named $Guid. Should they be consistent? Getting the value from the settings export object seems like a roundabout way to get it.

`$ReportUuid = $SettingsExportObject.report_uuid`

tkol2022:
### Testing long paths
This section covers testing with some long file names and folder paths. The purpose was so simulate the conditions where the path of the output JSON file exceeds the maximum system limit and triggers our newly added custom error message in the Merge-JsonOutput function. In some cases I am not getting the expected custom defined error message. 

**Case 1: long OutJsonFileName parameter value**
**I am not getting the expected custom error** because the GetFullPath cmdlet is failing before it gets to our custom error message.

`Invoke-Scuba -ProductNames sharepoint -CertificateThumbPrint "93fd41a761f434f12ee8ac3ff771347597435696" -AppID "db791fde-1ff0-493f-be3d-03d9ba6cb087" -Organization "y2zj1.onmicrosoft.com"  -NumberOfUUIDCharactersToTruncate 13 -OutJsonFileName "aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7eF8gH9iJ0kL1mN2oP3qR4sT5uV6wX7yZ8aB9cD0eF1gH2iJ3kL4mN5oP6qR7sT8uV9wX0yZ1aB2cD3eF4gH5iJ6kL7mN8oP9qR0sT1uV2wX3yZ4aB5cD6eF7gH8iJ9kL0mN1oP2qR3sT4uV5wX6yZ7aB8cD9eF0gH1iJ2kL3mN4oP5qR6sT7uV8wX9yZ0"`

image

**Case 2: long OutPath parameter value**
**This works as expected** because I get the custom error message in Merge-JsonOutput.

`Invoke-Scuba -ProductNames sharepoint -CertificateThumbPrint "93fd41a761f434f12ee8ac3ff771347597435696" -AppID "db791fde-1ff0-493f-be3d-03d9ba6cb087" -Organization "y2zj1.onmicrosoft.com"  -NumberOfUUIDCharactersToTruncate 13 -OutPath "C:\Users\tkolovos\Downloads\ScubaLongFolder\aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7eF8gH9iJ0kL1mN2oP3qR4sT5uV6wX7yZ8aB9cD0eF1gH2iJ3kL4mN5oP6qR7sT8uV9wX0yZ1aB2cD3eF4gH5iJ6kL7mN8oP9qR0sT1uV2wX3yZ4aB5cD6eF7gH8iJ9kL0mN1oP2qR3sT4uV5wX6yZ7aB8cD9eF"`

image

**Case 3: long OutFolderName parameter value**
**This works as expected** because I get the custom error message in Merge-JsonOutput.

`Invoke-Scuba -ProductNames sharepoint -CertificateThumbPrint "93fd41a761f434f12ee8ac3ff771347597435696" -AppID "db791fde-1ff0-493f-be3d-03d9ba6cb087" -Organization "y2zj1.onmicrosoft.com"  -NumberOfUUIDCharactersToTruncate 13 -OutFolderName "aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7eF8gH9iJ0kL1mN2oP3qR4sT5uV6wX7yZ8aB9cD0eF1gH2iJ3kL4mN5oP6qR7sT8uV9wX0yZ1aB2cD3eF4gH5iJ6kL7mN8oP9qR0sT1uV2wX3yZ4aB5cD6eF7gH8iJ9kL0mN1oP2qR3sT4uV5wX6yZ7aB8cD9eF"`

image

**Case 4: long OutFolderName parameter value that is exactly 256 characters (under the limit)**
**This works as expected** and it created the json merged in the long output folder.

`Invoke-Scuba -ProductNames sharepoint -CertificateThumbPrint "93fd41a761f434f12ee8ac3ff771347597435696" -AppID "db791fde-1ff0-493f-be3d-03d9ba6cb087" -Organization "y2zj1.onmicrosoft.com"  -NumberOfUUIDCharactersToTruncate 13 -OutFolderName "aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7eF8gH9iJ0kL1mN2oP3qR4sT5uV6wX7yZ8aB9cD0eF1gH2iJ3kL4mN5oP6qR7sT8uV9wX0yZ1aB2cD3eF4gH5iJ6kL7"`

image

**Case 5: long OutFolderName parameter value that is exactly 256 characters (under the limit) but no truncated characters**
**This works as expected** because I get the custom error message since the addition of the full GUID make the path exceed the limit.

`Invoke-Scuba -ProductNames sharepoint -CertificateThumbPrint "93fd41a761f434f12ee8ac3ff771347597435696" -AppID "db791fde-1ff0-493f-be3d-03d9ba6cb087" -Organization "y2zj1.onmicrosoft.com"  -NumberOfUUIDCharactersToTruncate 0 -OutFolderName "aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7eF8gH9iJ0kL1mN2oP3qR4sT5uV6wX7yZ8aB9cD0eF1gH2iJ3kL4mN5oP6qR7sT8uV9wX0yZ1aB2cD3eF4gH5iJ6kL7"`

image

adhilto:
> > We might want to do some data validation on the OutJsonFileName parameter. If you pass "*" as the name, it crashes. Maybe add a ValidatePattern to the parameter definition in Invoke-Scuba?
> > ```
> > [Parameter(Mandatory = $false, ParameterSetName = 'Configuration')]
> > [Parameter(Mandatory = $false, ParameterSetName = 'Report')]
> > [ValidateNotNullOrEmpty()]
> > [ValidatePattern('^[a-zA-Z0-9]+$')]
> > [string]
> > $OutJsonFileName = [ScubaConfig]::ScubaDefault('DefaultOutJsonFileName')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Again this is an enhancement I think we can make in the future but wanted to discuss.
> 
> Oh we should probably start checking for this for every file name parameter 👀. Will queue up an issue for this for every file name name parameter.
> 
> For this PR adding the validation to `OutJSONFileName` is a low lift.

I agree with adding validation, but I think it's not as trivial to implement as it seems. See https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names.

For example, `S{c}u b-a_(R)e%s^ult$s.json` is a completely valid file name. Will anyone actually use all those characters in a file name? I sure hope not, but we shouldn't impose stricter rules than Windows itself imposes. As some less extreme examples, the example regex in your comment wouldn't allow names such as "Scuba-Results.json", "Scuba_Results.json", and "Scuba Results.json".

TLDR: I think we should save this for a later issue to make sure we think through all the edge cases.

adhilto:
> I haven't tested this section of the code yet but I have a question. Does this code make the logic windows specific? What types of tweaks will we need in the future if we work on a cross-platform version that doesn't have the same file path limitation coded here?
> 
> image

Good point, this is a Windows specific limitation. Linux, for example, appears to impose a 255 character limit on file name but no limit on path. https://stackoverflow.com/questions/6571435/limit-on-file-name-length-in-bash

Perhaps a more simple (and more system agnostic) approach would be to simply remove all this added logic and just wrap the `Set-Content` call in a try/catch? Effectively let the system decide what it can or can't handle.

adhilto:
> > The probability of New-Guid generating an exception is probably negligibly low, but if it does fail and the 0000 value is created as the GUID, won't this cause filename collisions in the downstream file system where the results will be posted if another user happened to get an exception as well? There would be two files, both with the 0000 GUID in the filename. Again the probability is low.
> 
> I've been thinking about this as well. Should we just throw an full exception in this case and stop ScubaGear execution? If `New-Guid` does fail, there most likely is something catastrophically wrong with the underlying system.

I'd be ok with New-Guid failing being a fatal error.

mitchelbaker-cisa:
> > > The probability of New-Guid generating an exception is probably negligibly low, but if it does fail and the 0000 value is created as the GUID, won't this cause filename collisions in the downstream file system where the results will be posted if another user happened to get an exception as well? There would be two files, both with the 0000 GUID in the filename. Again the probability is low.
> > 
> > 
> > I've been thinking about this as well. Should we just throw an full exception in this case and stop ScubaGear execution? If `New-Guid` does fail, there most likely is something catastrophically wrong with the underlying system.
> 
> I'd be ok with New-Guid failing being a fatal error.

Capturing the two other locations in InvokeSCuBACached where the same guid logic is performed, fatal error should be thrown for these instances as well.

Also, since this try/catch block is repeated 3 times in the orchestrator, does it make sense to break it out as its own utility function? 

Screenshot (339)

adhilto:
> I see that we are truncating the filename GUID in multiple places, ConvertTo-ResultsCsv and Merge-JsonOutput resulting in duplicate logic that we have to keep in synch over time. What if we truncate the GUID in a helper function? Is it worth it?
> 
> image
> 
> image
> 
> Also, the Merge-JsonOutput function gets the GUID from the line of code below which gets it from the settings export object, whereas the ConvertTo-ResultsCsv function gets the GUID from a parameter named $Guid. Should they be consistent? Getting the value from the settings export object seems like a roundabout way to get it.
> 
> `$ReportUuid = $SettingsExportObject.report_uuid`

@tkol2022 I just made this refactor (addresses both comments here).

tkol2022:
> > > We might want to do some data validation on the OutJsonFileName parameter. If you pass "*" as the name, it crashes. Maybe add a ValidatePattern to the parameter definition in Invoke-Scuba?
> > > ```
> > > [Parameter(Mandatory = $false, ParameterSetName = 'Configuration')]
> > > [Parameter(Mandatory = $false, ParameterSetName = 'Report')]
> > > [ValidateNotNullOrEmpty()]
> > > [ValidatePattern('^[a-zA-Z0-9]+$')]
> > > [string]
> > > $OutJsonFileName = [ScubaConfig]::ScubaDefault('DefaultOutJsonFileName')
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Again this is an enhancement I think we can make in the future but wanted to discuss.
> > 
> > 
> > Oh we should probably start checking for this for every file name parameter 👀. Will queue up an issue for this for every file name name parameter.
> > For this PR adding the validation to `OutJSONFileName` is a low lift.
> 
> I agree with adding validation, but I think it's not as trivial to implement as it seems. See https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names.
> 
> For example, `S{c}u b-a_(R)e%s^ult$s.json` is a completely valid file name. Will anyone actually use all those characters in a file name? I sure hope not, but we shouldn't impose stricter rules than Windows itself imposes. As some less extreme examples, the example regex in your comment wouldn't allow names such as "Scuba-Results.json", "Scuba_Results.json", and "Scuba Results.json".
> 
> TLDR: I think we should save this for a later issue to make sure we think through all the edge cases.

Agree or honestly just let it fail whenever the orchestrator tries to create the file and avoid trying to be too fancy with validation logic, that like you said, might be non trivial. Also keep in mind, we need to think cross-platform.

tkol2022:
> > The probability of New-Guid generating an exception is probably negligibly low, but if it does fail and the 0000 value is created as the GUID, won't this cause filename collisions in the downstream file system where the results will be posted if another user happened to get an exception as well? There would be two files, both with the 0000 GUID in the filename. Again the probability is low.
> 
> I've been thinking about this as well. Should we just throw an full exception in this case and stop ScubaGear execution? If `New-Guid` does fail, there most likely is something catastrophically wrong with the underlying system.

Yes, I am not a fan of the 000 GUID. That just seems a little whacky although it will likely work :)

tkol2022:
> > I haven't tested this section of the code yet but I have a question. Does this code make the logic windows specific? What types of tweaks will we need in the future if we work on a cross-platform version that doesn't have the same file path limitation coded here?
> > image
> 
> Good point, this is a Windows specific limitation. Linux, for example, appears to impose a 255 character limit on file name but no limit on path. https://stackoverflow.com/questions/6571435/limit-on-file-name-length-in-bash
> 
> Perhaps a more simple (and more system agnostic) approach would be to simply remove all this added logic and just wrap the `Set-Content` call in a try/catch? Effectively let the system decide what it can or can't handle.

My preference would be to do this (your try/catch idea). Maybe we can catch a specific exception that occurs when the path is too long and then print the custom error message.

docs/configuration/parameters.md
```diff
@@ -160,21 +178,32 @@ The list of acceptable values are:
...
-| Default     | n/a    |
-| Config File | No     |  
+**NumberOfUUIDCharactersToTruncate** controls how many characters will be truncated from the report UUID when appended to the end of **OutJsonFileName**.
+
+| Parameter   | Value              |
+|-------------|--------------------|
+| Optional    | Yes                |
+| Datatype    | Integer            |
+| Default     | 18                 |
+| Config File | Yes                |
```
mitchelbaker-cisa, line 191: Add the default value for NumberOfUUIDCharactersToTruncate into full_config.yaml or some other sample config file.

mitchelbaker-cisa (changes requested):
- PS unit tests pass

- Ran `Invoke-SCuBA` against GCC tenant successfully
- Ran `Invoke-SCuBA -KeepIndividualJSON` against GCC tenant successfully, ProviderSettingsExport.json is generated
- Ran `Invoke-SCuBACached` against GCC successfully

- Ran `Invoke-SCuBA` against GCC high tenant successfully 
- Ran `Invoke-SCuBA -KeepIndividualJSON` against GCC high tenant successfully, ProviderSettingsExport.json is generated

Ran `Invoke-SCuBACached` with input that had multiple ScubaResults.json files; pulled from the most recent created
Ran Invoke-SCuBA`/`Invoke-SCuBACached` with non-default OutPath/OutJsonFileName parameters

Tested -NumberOfUUIDCharactersToTruncate in config file; works as expected with various cases (valid positive integer, invalid negative integer) although for invalid cases the error is caught in the generic MergeJson block as opposed to the standard message, "Hey, the integer you provided is not included 0, 13, 18, 36 provided in ValidateSet()."
Screenshot (341)

PowerShell/ScubaGear/Modules/Orchestrator.psm1
```diff
@@ -216,6 +219,7 @@ function Invoke-SCuBA {
         [Parameter(Mandatory = $false, ParameterSetName = 'Configuration')]
         [Parameter(Mandatory = $false, ParameterSetName = 'Report')]
         [ValidateNotNullOrEmpty()]
+        # [ValidatePattern('^[a-zA-Z0-9]+$')]
```
mitchelbaker-cisa, line 222: Remove comment, address in future issue 
```suggestion
```

PowerShell/ScubaGear/Modules/Orchestrator.psm1
```diff
@@ -374,22 +386,27 @@ function Invoke-SCuBA {
         # Tenant Metadata for the Report
         $TenantDetails = Get-TenantDetail -ProductNames $ScubaConfig.ProductNames -M365Environment $ScubaConfig.M365Environment
 
+        # Generate the GUID for the JSON
+        # TODO Stick the GUID within the ScubaConfig object to clean up parameter bloat
```
mitchelbaker-cisa, line 390: Is this comment still relevant? If not remove

PowerShell/ScubaGear/Modules/Orchestrator.psm1
mitchelbaker-cisa, line 222: Remove comment, address in future issue 
```suggestion
```

tkol2022:
We probably want to remove the explicit inclusion of the stacktrace in the future when Alex's error message handling change goes into effect because this may result in the stacktrace printing twice which would not be user friendly. Thoughts?

image

adhilto:
> We probably want to remove the explicit inclusion of the stacktrace in the future when Alex's error message handling change goes into effect because this may result in the stacktrace printing twice which would not be user friendly. Thoughts?
> 
> image

Good catch. I put that in for debugging, I hadn't realized I had pushed it.

tkol2022:
### Interesting result when Windows has long paths enabled
*Putting this here for future reference.

There is a Windows configuration that allows long paths (i.e. past the default 256 character size limit) when the registry value is configured as per below. My test system has long paths enabled so ScubaGear was able to run fine even past 256 characters.
HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem > LongPathsEnabled = 1

The Scuba results json file generated below has a total path length of 284 characters.

image

tkol2022:
### Results when testing ScubaGear inside a long path
I didn't get a chance to test this specific test case yesterday so I gave it a whirl today. This test simulates how ScubaGear behaves when executing inside a folder that is part of a very long path. The idea was to try and simulate the condition where the creation of the ScubaResults.json file results in an exception due to exceeding the 256 character Windows path limit. The test system has the Windows long paths disabled.

tldr: There is no action for us I don't think. Leave the code as-is. If I am missing something and you want me to re-test, let me know.

**Summary:**
Unfortunately, or fortunately (depending on your sense of humor), I was not able to verify that ScubaGear will catch the exception that occurs when the results json file path exceeds the character limit because I discovered that ScubaGear fails before it even reaches the exception catch logic for the json file. See screenshot for reference. Apparently the log path causes other components to fail first.

ScubaGear was placed into the following folder (194 characters in length):
C:\Users\adminz\Downloads\LongPath\aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7eF8gH9iJ0kL1mN2oP3qR4sT5uV6wX7yZ8aB9cD0eF1gH2iJ3kL4mN5oP6qR7sT8uV9wX0yZ1aB2cD3123456789123456789123456789123456789

image

tkol2022:
### Testing Long Path cases from yesterday (regression)
This is a summary of doing a regression of the long path cases that I ran yesterday. Each test executed below produced a failure in different parts of the code. I'm not if it makes sense to generalize here but I'm thinking that ScubaGear will probably fail in other parts of the code before it arrives at any new logic added in this PR to catch path length errors in Merge-JsonOutput.
I don't think we should change the code at this point. Keep the cases below for future reference.

**Case 1: long OutJsonFileName parameter value**
ScubaGear fails during json reports aggregation.
image


**Case 2: long OutPath parameter value**
ScubaGear fails before it even gets to the results JSON.

image


**Case 3: long OutFolderName parameter value**
ScubaGear fails before it even gets to the results JSON.

image


**Case 4: long OutFolderName parameter value that is exactly 256 characters (under the limit)**
ScubaGear fails before it gets to the results JSON. Depending on the length of the OutFolderName it can fail in multiple places including during the report creation.

image

tkol2022:
### Regression test summary

- [ ] Test all possible GUID size values for the NumberOfUUIDCharactersToTruncate parameter
- [ ] Test all the cases documented in the PR again 
- [ ] Test the functional test orchestrator
- [ ] Test running scubagear inside a long path without any funny parameters
- [ ] Test the manual run cached

docs/configuration/parameters.md
```diff
@@ -274,7 +274,7 @@ Invoke-SCuBA -ProductNames teams `
 
 ## OutJsonFileName
 
-**OutJsonFileName** renames the uber output JSON file that is created after a ScubaGear run. This should only be the base file name, as the extension `.json` will automatically be added. 
+**OutJsonFileName** specifies the base file name of the uber output JSON file that is created after a ScubaGear run. This should only be the base file name; the report UUID as well as the extension, `.json`, will automatically be added.
```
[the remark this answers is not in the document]
tkol2022: Resolved. Comment was modified.

tkol2022 (approved):
Nice work.

Make sure to run the functional tests action.

mitchelbaker-cisa (approved):

PR #2186, opened 2026-06-01
fix: add OrgName and OrgUnitName to top-level Metadata in ScubaResults.json

### LINKED ISSUE #2137: Add OrgName and OrgUnitName to the top level Metadata key in ScubaResults.json
### Prerequisites

### 💡 Summary

This issue is to replicate the  `OrgName`, `OrgUnitName` to the top level `Metadata` key in the `ScubaResults_*.json`. Originally they were supposed to be added there by default from the config file. This issue is to clean up that loose end. 

Note that downstream processes rely on these values being at these locations: `Raw.scuba_config.OrgName` and `Raw.scuba_config.OrgUnitName`. 
The new replicated `Metadata.OrgName` and `Metadata.OrgUnitName` values are purely for faster file identification. 
 
If the values are present. Add their values to corresponding `OrgName` and `OrgUnitName` keys under Metadata.
Otherwise, keep the keys present in `Metadata` but leave them `null`

1. For example if the `OrgName` and `OrgUnitName`  was specified in the config file. Then the metadata should look something like this. 

Image

---------------------------------------------------------------------------------------------------------------

2. Otherwise the values will appear like this if `OrgName` and `OrgUnitName` are not specified in the config file. 

Image

---------------------------------------------------------------------------------------------------------------

3. Currently the keys are located at `Raw.scuba_config.OrgName` and `Raw.scuba_config.OrgUnitName`
within a `ScubaResults_*.json`
Image

### Motivation and context

Cleaning up a loose end.  Faster file identification for analytics.

### Implementation notes

1. Requires modifying `Merge-JSONOutput` within the `Orchestrator.psm1` to add the new keys and values. 

### Acceptance criteria

The values are present in the JSON like so.  
1. Values present within config file Image
2. Values not present in the config file Image

stark256-spec:
Closes #2137

## Summary

Replicates `OrgName` and `OrgUnitName` from `Raw.scuba_config` into the top-level `Metadata` key of `ScubaResults_*.json`, so downstream analytics processes can identify a results file without traversing the `Raw` object.

**`Orchestrator.psm1` — `Merge-JsonOutput`**

Added two fields to the `$MetaData` object built at the start of result aggregation:

```powershell
"OrgName"     = $SettingsExportObject.scuba_config.OrgName;
"OrgUnitName" = $SettingsExportObject.scuba_config.OrgUnitName;
```

`$SettingsExportObject` is the already-loaded provider JSON, so no additional I/O is needed. Both keys are always present in `Metadata`; when `OrgName`/`OrgUnitName` are not set in the config file the values are `null`, matching the accepted design in the issue.

**`Merge-JsonOutput.Tests.ps1`**

- Updated the existing `ConvertFrom-Json` mock to include `scuba_config` with `OrgName`/`OrgUnitName`
- Added a context block for the case where both values are populated
- Added a context block for the case where `scuba_config` is empty (null values)

## Test plan

- [x] `Invoke-Pester PowerShell/ScubaGear/Testing/Unit/PowerShell/Orchestrator/Merge-JsonOutput.Tests.ps1`
- [x] Verify `ScubaResults_*.json` contains `Metadata.OrgName` and `Metadata.OrgUnitName` after a full run with `OrgName`/`OrgUnitName` set in config
- [x] Verify the keys are present but `null` when not set in config

### DIFF
--- a/PowerShell/ScubaGear/Modules/Orchestrator.psm1
+++ b/PowerShell/ScubaGear/Modules/Orchestrator.psm1
@@ -1373,6 +1373,8 @@ function Merge-JsonOutput {
                 "ToolVersion" = $ModuleVersion;
                 "TimestampZulu" = $TimestampZulu;
                 "ReportUUID" = $Guid;
+                "OrgName" = $SettingsExportObject.scuba_config.OrgName;
+                "OrgUnitName" = $SettingsExportObject.scuba_config.OrgUnitName;
             }
 
 
--- a/PowerShell/ScubaGear/Modules/Orchestrator.psm1
+++ b/PowerShell/ScubaGear/Modules/Orchestrator.psm1
@@ -1366,6 +1366,8 @@ function Merge-JsonOutput {
                 "TenantId" = $TenantDetails.TenantId;
                 "DisplayName" = $TenantDetails.DisplayName;
                 "DomainName" = $TenantDetails.DomainName;
+                "OrgName" = $SettingsExportObject.scuba_config.OrgName;
+                "OrgUnitName" = $SettingsExportObject.scuba_config.OrgUnitName;
                 "ProductSuite" = "Microsoft 365";
                 "ProductsAssessed" = $FullNames;
                 "ProductAbbreviationMapping" = $ProductAbbreviationMapping
@@ -1373,8 +1375,6 @@ function Merge-JsonOutput {
                 "ToolVersion" = $ModuleVersion;
                 "TimestampZulu" = $TimestampZulu;
                 "ReportUUID" = $Guid;
-                "OrgName" = $SettingsExportObject.scuba_config.OrgName;
-                "OrgUnitName" = $SettingsExportObject.scuba_config.OrgUnitName;
             }
 
 
--- a/PowerShell/ScubaGear/Testing/Unit/PowerShell/Orchestrator/Merge-JsonOutput.Tests.ps1
+++ b/PowerShell/ScubaGear/Testing/Unit/PowerShell/Orchestrator/Merge-JsonOutput.Tests.ps1
@@ -13,6 +13,10 @@ InModuleScope Orchestrator {
                     "Results"        = @();
                     "timestamp_zulu" = "";
                     "report_uuid" = "00000000-0000-0000-0000-000000000000"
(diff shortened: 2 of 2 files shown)


buidav:
@stark256-spec Thanks for taking on the feature request! 
We'll take a look and get back to you on approval/requests as soon as we can.

PowerShell/ScubaGear/Modules/Orchestrator.psm1
```diff
@@ -1373,6 +1373,8 @@ function Merge-JsonOutput {
                 "ToolVersion" = $ModuleVersion;
                 "TimestampZulu" = $TimestampZulu;
                 "ReportUUID" = $Guid;
+                "OrgName" = $SettingsExportObject.scuba_config.OrgName;
```
DickTracyII: I tested this in different tenants and it is coming back with the added metadata. mM only suggestion is to move the `OrgName` and `OrgUnitName` higher up under `DomainName` for easier find. 

Image

DickTracyII (approved):
It does include the new meta items. good work.

PowerShell/ScubaGear/Modules/Orchestrator.psm1
stark256-spec: Thanks for testing across tenants! Moved `OrgName` and `OrgUnitName` to sit directly under `DomainName` in the metadata object.

buidav (approved):
Scenarios tested. 

- Running ScubaGear without a config file 
  - Result: OrgName and OrgUnitName are `null` within `Metadata`
- Running ScubaGear with a config file where the OrgName and OrgUnitName are not specified as keys 
  - Result: OrgName and OrgUnitName are `null` within `Metadata`
- Running ScubaGear with a config file where the OrgName is specified but OrgUnitName is not specified
  - Result: OrgName contains expected string value but OrgUnitName is `null` within `Metadata`
 - Running ScubaGear with a config file where the OrgName is not specified but OrgUnitName is specified
   - Result: OrgName is null but OrgUnitName contains expected string value within `Metadata`
 - Running ScubaGear with a config file where both OrgName and OrgUnitName is specified
   - Result: OrgName and OrgUnitName both contain expected string values within `Metadata`

Thanks for taking on the issue!

PR #2223, opened 2026-06-12
Fix DMARC discovery with RFC 9989 DNS tree walk

### LINKED ISSUE #89: DMARC Edge Case Ignored
There is a quirk in the DMARC protocol which we left as out-of-scope for the initial proof of concept code that should be handled at some point.

DMARC records can be stored either under the full domain name (e.g., `test365.cisa.dhs.gov`), or under the "organizational domain," which is basically the second-level domain plus the tld (e.g., `dhs.gov`). Our code currently handles everything I've described up to this point.

The edge case that we aren't handling is that technically there exist "tlds" that in fact consist of multiple labels, mostly because of country-code tlds, e.g., fed.us (see [here](https://publicsuffix.org/list/public_suffix_list.dat) for a list maintained by Mozilla). So if the domain we're trying to find the DMARC record for were `subdomain.example.fed.us`, we would need to check for the DMARC record under two places:
1. subdomain.example.fed.us
2. example.fed.us

Our code currently would check for:
1. subdomain.example.fed.us
2. **fed.us**

A description of the proper way to do this: https://datatracker.ietf.org/doc/html/rfc7489#section-3.2

omobolajiadeyan:
## Summary

Replace the fixed last-two-label DMARC fallback with the bounded DNS Tree Walk defined by RFC 9989 section 4.10.

## Changes

- query the full domain first;
- when no direct policy exists, walk toward the DNS root and retain the highest valid DMARC policy found;
- cap discovery at eight total DNS queries, skipping intermediate labels for deeply nested domains as required by RFC 9989;
- stop early at an explicit `psd=y` or `psd=n` boundary;
- require exactly one answer beginning with `v=DMARC1;` before treating a response as a DMARC policy;
- preserve log entries from every attempted DNS query; and
- remove the duplicate domain-label split in the previous fallback.

## Tests

Added regression coverage for:

- direct-domain policy lookup;
- highest valid policy selection during a normal tree walk;
- `subdomain.example.fed.us` resolving through `_dmarc.example.fed.us`;
- stopping at a PSD boundary; and
- the eight-query limit for a deeply nested domain.

Validation:

- focused DMARC Pester tests: 6 passed;
- complete EXO provider Pester folder: 42 passed;
- `git diff --check`: passed;
- direct PSScriptAnalyzer on changed files: no new findings in changed code (one existing `PSUseSingularNouns` warning remains at `ExportEXOProvider.psm1:299`).

Fixes #89

Standard: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
@@ -717,26 +717,51 @@ function Get-ScubaDmarcRecord {
             continue
         }
         $LogEntries = @()
-        # First check to see if the record is available at the full domain level
         $DomainName = $d.DomainName
-        $Response = Invoke-RobustDnsTxt "_dmarc.$DomainName" -PreferredDnsResolvers $PreferredDnsResolvers `
-            -SkipDoH $SkipDoH
-        $LogEntries += $Response.LogEntries
-        if ($Response.Answers.Length -eq 0) {
-            # The domain does not exist. If the record is not available at the full domain
-            # level, we need to check at the organizational domain level.
-            $Labels = $d.DomainName.Split(".")
-            $Labels = $d.DomainName.Split(".")
-            $OrgDomain = $Labels[-2] + "." + $Labels[-1]
-            $Response = Invoke-RobustDnsTxt "_dmarc.$OrgDomain" -PreferredDnsResolvers $PreferredDnsResolvers `
-                -SkipDoH $SkipDoH
-            $LogEntries += $Response.LogEntries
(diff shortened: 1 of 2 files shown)


omobolajiadeyan:
The implementation is ready for review. The three workflow runs are currently `action_required` because this is a first-time fork contribution. When convenient, please approve the CI, Smoke Test, and Workflow Pipeline runs. Local validation passed: 6 focused RFC 9989 DMARC tests and all 42 EXOProvider tests.

omobolajiadeyan:
This PR was closed when I mistakenly removed its source fork during repository cleanup; it was not closed because of a review decision. Since the RFC 9989 direction had been confirmed in issue #89, I still have the tested implementation locally. I will not recreate the PR unless the maintainers would still like it resubmitted. Apologies for disrupting the review history.

Issue #2219, opened 2026-06-10
Improve design of EXO policies that call DNS services due to performance problems in some environments


tkol2022:
### Prerequisites

- [x] This issue has an informative and human-readable title.

### 💡 Summary

In some environments, such as running from my laptop against our G5 tenant, the EXO export provider has noticeable performance problems which are related to the DNS-specific functions: Get-ScubaSpfRecord, Get-ScubaDkimRecord, Get-ScubaDmarcRecord. To run these three functions it took almost 4 minutes during an execution this morning.

This problem seems to be related to specific environments, perhaps based on their DNS configuration. 

<img width="1111" height="454" alt="Image" src="https://github.com/user-attachments/assets/3aa0dd33-06b1-4451-9f65-a7b37c67e699" />

### Motivation and context

To improve the user experience, it is always our goal to make ScubaGear run with better performance.

### Implementation notes

1. Determine the root cause of the problem. I will be happy to help diagnose since it seems to be happening consistently in my execution environment.
2. Investigate potential alternative approaches to getting the DNS information.
3. Prototype a solution. If it works, open a PR.

### Acceptance criteria

- [ ] Developer has completed the items above and testing their solution.

PR #2237, opened 2026-06-20
Fix DMARC policy discovery tree walk

omobolajiadeyan:
## Intent

This PR updates `Get-ScubaDmarcRecord` so DMARC fallback discovery follows the bounded DNS Tree Walk described in RFC 9989 instead of assuming the applicable policy is always at the final two labels of the domain.

## Behavior being fixed
Proposed solution for https://github.com/cisagov/ScubaGear/issues/89.

The current fallback builds the policy domain as `labels[-2] + "." + labels[-1]`. That works for simple domains such as `sub.example.com` -> `example.com`, but it can miss the correct DMARC policy domain when the domain has a multi-label public suffix or otherwise needs parent-domain walking.

Example:

- Input author domain: `subdomain.example.fed.us`
- Current final-two-label fallback checks: `_dmarc.fed.us`
- Expected tree-walk behavior checks parent domains and can find: `_dmarc.example.fed.us`

## What changed

- Keep the existing author-domain lookup first.
- If no DMARC record is found there, walk parent domains using a bounded DNS Tree Walk.
- Limit total DNS TXT lookups to eight.
- Preserve the existing output shape and logging behavior.
- Add Pester tests for parent-domain discovery, multi-label public suffix behavior, and the eight-query limit.

## Reference

- RFC 9989 Section 4.10: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10

## Verification

```powershell
Import-Module Pester -RequiredVersion 5.7.1
Invoke-Pester -Path 'PowerShell\ScubaGear\Testing\Unit\PowerShell\Providers\EXOProvider\Get-ScubaDmarcRecord.Tests.ps1' -Output Detailed
```

Result: 12 tests passed, 0 failed.

### DIFF
--- a/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
@@ -741,14 +741,36 @@ function Get-ScubaDmarcRecord {
             -SkipDoH $SkipDoH
         $LogEntries += $Response.LogEntries
         if ($Response.Answers.Length -eq 0) {
-            # The domain does not exist. If the record is not available at the full domain
-            # level, we need to check at the organizational domain level.
-            $Labels = $d.DomainName.Split(".")
-            $Labels = $d.DomainName.Split(".")
-            $OrgDomain = $Labels[-2] + "." + $Labels[-1]
-            $Response = Invoke-RobustDnsTxt "_dmarc.$OrgDomain" -PreferredDnsResolvers $PreferredDnsResolvers `
-                -SkipDoH $SkipDoH
-            $LogEntries += $Response.LogEntries
+            # RFC 9989 replaces Public Suffix List organizational-domain discovery with
+            # a bounded DNS Tree Walk. Keep the author-domain query above, then walk
+            # parent domains without exceeding eight total DNS queries.
+            $Labels = @($d.DomainName.Split(".") | Where-Object { $_ -ne "" })
+            $Queries = 1
+            $CandidateResponse = $null
+
+            while ($Labels.Length -gt 1 -and $Queries -lt 8) {
+                if ($Labels.Length -ge 8) {
+                    $Labels = @($Labels[($Labels.Length - 7)..($Labels.Length - 1)])
+                }
+                else {
+                    $Labels = @($Labels[1..($Labels.Length - 1)])
+                }
+
+                $TreeWalkDomain = $Labels -join "."
+                $CandidateResponse = Invoke-RobustDnsTxt "_dmarc.$TreeWalkDomain" `
+                    -PreferredDnsResolvers $PreferredDnsResolvers `
+                    -SkipDoH $SkipDoH
+                $Queries += 1
+                $LogEntries += $CandidateResponse.LogEntries
+
+                if ($CandidateResponse.Answers.Length -gt 0) {
+                    $Response = $CandidateResponse
+                }
+            }
+
+            if ($Response.Answers.Length -eq 0 -and $null -ne $CandidateResponse) {
+                $Response = $CandidateResponse
+            }
         }
 
         $DomainName = $d.DomainName
--- a/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
@@ -763,13 +763,27 @@ function Get-ScubaDmarcRecord {
                 $Queries += 1
                 $LogEntries += $CandidateResponse.LogEntries
 
-                if ($CandidateResponse.Answers.Length -gt 0) {
-                    $Response = $CandidateResponse
-                }
-            }
+                # RFC 9989 tree-walk steps 2 and 6: an answer only counts as this
+                # target's DMARC policy if it is itself a valid DMARC record (begins
+                # with the version tag). Multiple valid records at the same target are
+                # ambiguous per RFC 7489 6.6.3 and are treated as no usable record
+                # there, same as zero answers. A single valid record carrying the psd
+                # tag marks the walk boundary explicitly, so stop there instead of
+                # continuing unconditionally through the remaining labels.
+                $ValidAnswers = @($CandidateResponse.Answers | Where-Object { $_ -match '^v=DMARC1' })
 
-            if ($Response.Answers.Length -eq 0 -and $null -ne $CandidateResponse) {
-                $Response = $CandidateResponse
+                if ($ValidAnswers.Length -eq 1) {
+                    $Response = [PSCustomObject]@{
+                        "Answers"    = @($ValidAnswers)
+                        "Errors"     = $CandidateResponse.Errors
+                        "NXDomain"   = $CandidateResponse.NXDomain
+                        "LogEntries" = $CandidateResponse.LogEntries
+                    }
+
+                    if ($ValidAnswers[0] -match 'psd=(n|y)\b') {
+                        break
+                    }
+                }
             }
         }
 
--- a/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
+++ b/PowerShell/ScubaGear/Modules/Providers/ExportEXOProvider.psm1
@@ -740,7 +740,27 @@ function Get-ScubaDmarcRecord {
         $Response = Invoke-RobustDnsTxt "_dmarc.$DomainName" -PreferredDnsResolvers $PreferredDnsResolvers `
             -SkipDoH $SkipDoH
         $LogEntries += $Response.LogEntries
-        if ($Response.Answers.Length -eq 0) {
+        # RFC 9989 step 2 applies to the starting target as well as every
+        # subsequent tree-walk target. Do not let unrelated TXT records or an
+        # ambiguous set of DMARC records suppress parent-domain discovery.
+        $ValidAnswers = @($Response.Answers | Where-Object { $_ -match '^v=DMARC1' })
+
+        if ($ValidAnswers.Length -eq 1) {
+            $Response = [PSCustomObject]@{
+                "Answers"    = @($ValidAnswers)
+                "Errors"     = $Response.Errors
+                "NXDomain"   = $Response.NXDomain
+                "LogEntries" = $Response.LogEntries
+            }
+        }
+        else {
+            $Response = [PSCustomObject]@{
+                "Answers"    = @()
+                "Errors"     = $Response.Errors
+                "NXDomain"   = $Response.NXDomain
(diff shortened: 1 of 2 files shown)


FollyBeachGurl:
@omobolajiadeyan Hello. Can you briefly state the intent of this pull request and the specific issue or behavior you want us to fix?

omobolajiadeyan:
Thanks for asking. I updated the PR body to make the intent clearer.

The intent is to fix the DMARC fallback path in `Get-ScubaDmarcRecord`. Today, when no DMARC record is found at the full author domain, the code falls back to the final two labels of the domain. That works for simple domains such as `sub.example.com` -> `example.com`, but it can miss the applicable policy for multi-label public suffix cases.

Specific behavior this PR is trying to fix:

- Input author domain: `subdomain.example.fed.us`
- Current final-two-label fallback checks: `_dmarc.fed.us`
- Expected bounded tree-walk behavior can find: `_dmarc.example.fed.us`

The change keeps the existing author-domain lookup first, then performs a bounded parent-domain walk with an eight-query limit. The tests cover parent-domain discovery, the multi-label `example.fed.us` case, and the query limit.

omobolajiadeyan:
Rebased this branch onto the current upstream main and reran the focused tests with Pester 5.7.1. All 6 Get-ScubaDmarcRecord tests pass, including parent discovery, the multi-label public-suffix case, and the RFC 9989 eight-query limit. The PR is ready for review.

adhilto:
Hi @omobolajiadeyan, thanks for your contribution. There's one issue that needs to be resolved before this can be merged. Currently, your implementation will walk through the entire tree regardless of whether or not if gets an answer. We need to follow the algorithm described in section [4.10](https://datatracker.ietf.org/doc/html/rfc9989#section-4.10) of the RFC as closely as possible, which includes stopping once an answer is found. Specifically, steps 2 and 6 of the algorithm described there are missing here.

omobolajiadeyan:
Thanks, @adhilto — you’re right that the current implementation is missing the record evaluation and stop conditions from steps 2 and 6.

I’ll update the walk so that, after each lookup, it discards records that do not begin with the current DMARC `v` tag, treats multiple DMARC policy records at one target as unusable, and stops when the single remaining valid record contains `psd=n` or `psd=y`, instead of continuing unconditionally through the remaining labels. I’ll add focused Pester coverage proving both the early-stop behavior and continued walking when those stop conditions are not met, then rerun the full `Get-ScubaDmarcRecord` test file and post the results here.

Thanks for pointing to the exact RFC steps.

omobolajiadeyan:
Pushed the fix I described above. The tree walk now:

- Discards any answer at a `_dmarc.<name>` target that doesn't begin with the `v=DMARC1` tag, so a non-DMARC TXT record there is no longer treated as that domain's policy.
- Treats multiple valid DMARC records at the same target as unusable (RFC 7489 6.6.3 discovery-failure case) rather than picking one.
- Stops the walk as soon as a single valid record carries `psd=n` or `psd=y`, instead of continuing unconditionally through the remaining bounded labels.

Added four new Pester cases in `Get-ScubaDmarcRecord.Tests.ps1` covering both stop conditions (`psd=y`, `psd=n`) and both discard conditions (non-DMARC answer, ambiguous multiple answers) — each also asserts the walk correctly continues past the discarded target. The existing lookup and eight-query-cap tests are unaffected since neither exercises a `psd` tag.

omobolajiadeyan:
Addressed the remaining RFC 9989 starting-point gap. The author-domain response now goes through the same version-tag and ambiguity validation as every parent target, so unrelated TXT data or multiple DMARC records cannot suppress the tree walk. Added focused regressions for both cases; all 12 `Get-ScubaDmarcRecord` Pester tests pass locally.

omobolajiadeyan:
@adhilto The RFC 9989 feedback has been implemented: starting and parent responses are version-tag validated, ambiguous records are discarded, the walk honors the `psd` stop conditions, and total DNS queries remain capped at eight. Focused coverage now includes those cases, with all 12 `Get-ScubaDmarcRecord` tests passing locally. Could you please re-review and confirm that steps 2 and 6 are now satisfied?

adhilto:
Hi @omobolajiadeyan, sorry for the delay and thanks for your update. There are still some issues that need to be addressed. Per section 4.10.1, DMARC policy discovery should use the following preference:
- The Author Domain
- The Organizational Domain of the Author Domain
- The PSD of the Author Domain

However, with your current implementation, there are cases where the PSD will incorrectly take precedence over the organizational domain.

As a concrete example, consider the fake domain example.cisa.gov.

First query, `_dmarc.example.cisa.gov`, returns no answer:
```
> Resolve-DnsName _dmarc.example.cisa.gov txt

Name                        Type TTL   Section    PrimaryServer               NameAdministrator           SerialNumber
----                        ---- ---   -------    -------------               -----------------           ------------
cisa.gov                    SOA  1800  Authority  blue.foundationdns.com      dns.cloudflare.com          2411858014
```

Second query, `_dmarc.example.cisa.gov`, returns a valid DMARC record, but it does not include a `psd` tag so execution continues:
```
> Resolve-DnsName _dmarc.cisa.gov txt

Name                                     Type   TTL   Section    Strings
----                                     ----   ---   -------    -------
_dmarc.cisa.gov                          TXT    28800 Answer     {v=DMARC1; p=reject; pct=100;
                                                                 rua=mailto:DMARC@hq.dhs.gov,
                                                                 mailto:reports@dmarc.cyber.dhs.gov}
```

Third query, `_dmarc.gov`, also returns a valid DMARC record, but this one has `psd=y`.
```
> Resolve-DnsName _dmarc.gov txt

Name                                     Type   TTL   Section    Strings
----                                     ----   ---   -------    -------
_dmarc.gov                               TXT    60    Answer     {v=DMARC1; p=reject; sp=none; np=reject;
                                                                 rua=mailto:dotgov_dmarc@cisa.dhs.gov;
                                                                 psd=y}
```

Your current implementation will select the final DMARC record, the one corresponding to `_dmarc.gov`, as the DMARC policy to evaluate. However, it should actually choose the DMARC record corresponding to `_dmarc.cisa.gov`, because per section [4.10.2](https://datatracker.ietf.org/doc/html/rfc9989#section-4.10.2), the organizational domain is the domain one level below the domain with `psd=y`.

omobolajiadeyan:
Thanks, @adhilto. I’ve updated the implementation to preserve the RFC 9989 policy preference when the tree walk reaches a `psd=y` boundary.

The walk now determines the Organizational Domain as the domain exactly one label below the PSD. If a valid DMARC record was discovered at that Organizational Domain, that record is retained instead of being overwritten by the PSD record. The PSD record is used only when the Organizational Domain has no valid record.

I added focused regression coverage for both cases:

- `example.cisa.gov` has no author-domain record, `_dmarc.cisa.gov` has a valid policy, and `_dmarc.gov` has `psd=y`; the `cisa.gov` policy is selected.
- An intervening domain has a record but the actual Organizational Domain does not; the `psd=y` record remains the applicable fallback.

All 14 focused `Get-ScubaDmarcRecord` Pester tests pass. The fix is in commit `71eb9a1a`.

FollyBeachGurl:
Moving to draft while conversations continue.

omobolajiadeyan:
Revalidated the current head with Pester 5.7.1: all 14 focused Get-ScubaDmarcRecord tests pass, including Organizational Domain precedence over the PSD and PSD fallback when the Organizational Domain has no valid record. Please let me know which remaining concerns should be addressed while the pull request remains in draft.

adhilto (commented):
@omobolajiadeyan thanks for addressing my feedback. I have a few final comments for the team to consider, but your implementation looks solid to me.

After reviewing the code, I ran a few small tests and everything came out as expected. Then I ran a large test, with 282 domains. I also ran the same test on the main branch. Of those, a few of the results differed, but in all cases the difference was expected and accurate. They were all due to the presence of a DMARC record at the `.gov` TLD, which is now included in the DMARC policy discovery path. As a side note, this change does increase the priority of https://github.com/cisagov/ScubaGear/issues/2294, as the DMARC record at `.gov` has `p=reject` and `sp=none`, so all cases where ScubaGear relies on that DMARC record will be incorrectly reported as compliant by ScubaGear for MS.EXO.4.2.

I do want to note some metrics as this PR is relevant to https://github.com/cisagov/ScubaGear/issues/2219 (@tkol2022 and @atuomit for awareness). The updated RFC will result in more DNS queries. For my above test with 282 domains:
- On main, ScubaGear made 432 DNS queries relating to DMARC (on average, 1.5 queries per domain).
- On this branch, ScubaGear made 663 DNS queries relating to DMARC (on average, 2.4 queries per domain).
- On main, Get-ScubaDmarcRecord took 80 seconds.
- On this branch, Get-ScubaDmarcRecord took 131 seconds.

Obviously, most people don't have 200+ domains, but I think it'd be reasonable to say this change could make the DMARC portion take somewhere between 1.5 and 2 times longer. The SPF and DKIM portions are unaffected.

sylvesterkaczmarek (commented):
The new validity filter is prefix-based: ^v=DMARC1 also accepts malformed versions such as v=DMARC10; p=reject. DMARC requires the version value to be exactly DMARC1, so those records should be ignored and the policy walk should continue. Could this require the tag delimiter/end after DMARC1 and add a malformed-version regression?

omobolajiadeyan:
﻿Thanks @adhilto for the broader validation and the DNS-query/runtime measurements. The additional cost and the related subdomain-policy issue in #2294 are useful considerations for the team.

@sylvesterkaczmarek, confirmed: both the author-domain and parent-domain filters currently use `^v=DMARC1`, and that expression also matches `v=DMARC10; p=reject`. Your requested boundary check is needed in both places. Regression coverage should exercise a malformed version at the author domain and at a parent target, confirming discovery continues to a usable policy; the placeholder `v=DMARC1...` test data also needs to become properly delimited records. This correction is still outstanding; the existing test results do not cover it.

adhilto:
> The new validity filter is prefix-based: ^v=DMARC1 also accepts malformed versions such as v=DMARC10; p=reject. DMARC requires the version value to be exactly DMARC1, so those records should be ignored and the policy walk should continue. Could this require the tag delimiter/end after DMARC1 and add a malformed-version regression?

@sylvesterkaczmarek you are correct, but in interest of moving this PR along, I'm going to recommend opening a follow-up issue to address that.

@atuomit ready for your review

sylvesterkaczmarek:
Opened the requested follow-up for the DMARC version-tag boundary: https://github.com/cisagov/ScubaGear/issues/2412. No need to hold this PR on that point from me.

atuomit (approved):
Tested on all four tenants in interactive and non-interactive mode. Passes all functional and unit tests.

adhilto (approved):

Issue #2242, opened 2026-06-23
AAD report does not contain background color coding for policies when processing certain provider JSON files


tkol2022:
### Prerequisites

- [x] This issue has an informative and human-readable title.

### ScubaGear Version

main branch

### Operating System

windows 11

### PowerShell Version

5.1

### M365 Environment and License(s)

G3 but the bug may have nothing to do with the license and could be about other logic in the report generation.

### 🐛 Summary

When running Invoke-ScubaCached with a specific agency provider JSON file, I noticed that ScubaGear does not provide the background color coding for Pass (green) and Fail (red) policies. The report just has a white background.

### Steps to reproduce

1. Contact @tkol2022 to get a copy of a provider JSON file that reproduces this issue. There is something unique about this provider file that causes the issue.
2. Run the file using the command below and open the AAD report to observe the lack of background colors.

`Invoke-SCuBACached -Productnames aad -ExportProvider $false -OutPath C:\Users\tkolovos\Downloads\Scuba\Testing -OutProviderFileName ProviderSettingsExport -KeepIndividualJSON`

<img width="1178" height="518" alt="Image" src="https://github.com/user-attachments/assets/939a6aed-3fcc-4929-86da-529683621ee2" />

### Expected behavior

I expect the HTML report to have the proper background colors.

mitchelbaker-cisa:
Closed by #2258

PR #2258, opened 2026-07-01
Fix missing background color coding in AAD report when provider JSON lacks report data keys

stark256-spec:
## 🗣 Description

Proposed solution for: #2242

### 💭 Motivation and context

When `Invoke-SCuBACached` is run against a provider JSON file that lacks one of the AAD report data keys (`cap_table_data`, `risky_applications`, `risky_third_party_service_principals`, `severity_score_weights`) — e.g. an export created by an older ScubaGear version — the AAD report renders with no Pass/Fail background colors, as described in #2242.

Root cause (two layers):

1. **Report generation (PowerShell 5.1):** `CreateReport.psm1` calls `ConvertTo-Json $SettingsExport.risky_applications` etc. directly. On Windows PowerShell 5.1, `ConvertTo-Json $null` fails parameter binding ("Cannot bind argument to parameter 'InputObject' because it is null"). The error is statement-terminating only, so the report still generates — but the variable is left empty, producing an empty JSON data island: `<script type='application/json' id='risky-apps-json'>  </script>`. (PowerShell 7 returns the literal `null` here, which is why this only reproduces on 5.1 — the version reported in the issue.)

2. **Report rendering (JavaScript):** `IndividualReport.js` loads all JSON data islands at the top of the `DOMContentLoaded` handler with `getJsonData`, which throws on invalid JSON. `JSON.parse("")` throws, the exception aborts the whole handler, and `colorRows()` is never reached — so every policy table renders with a white background. The server-rendered tables are unaffected, matching the screenshot in the issue.

Note the unit test stub `ProviderSettingsExport.json` already reproduces the data shape that triggers this: it contains `cap_table_data` but none of the risky app keys.

### Changes

- `CreateReport.psm1`: fall back to the JSON literal `"null"` when a report data key is absent from the provider export, instead of passing `$null` to `ConvertTo-Json`. The report scripts already handle `null` islands (`buildExpandableTable` returns early).
- `Utils.js`: add `tryGetJsonData`, which returns `null` and logs to the console instead of throwing, so one bad data island cannot abort the rest of report rendering.
- `IndividualReport.js`: use `tryGetJsonData` for the data island loads, so `colorRows()` (and the rest of the handler) always runs.
- `New-Report.Tests.ps1`: assert that every JSON data island embedded in the generated AAD report is non-empty, valid JSON. Because the CI unit test job runs under Windows PowerShell 5.1 (`shell: powershell`), this test fails on `main` and passes with this fix.

### 🧪 Testing

- `Invoke-Pester PowerShell/ScubaGear/Testing/Unit/PowerShell/CreateReport/` — 73/73 passing
- `Invoke-PSSA` (repo PSScriptAnalyzer config) — 0 errors, 0 warnings
- Verified `tryGetJsonData` degrades gracefully for all three failure modes (empty island, missing element, `null` island) and that `buildExpandableTable(null, ...)` returns early as before

I did not have access to the specific agency provider file referenced in the issue, so it would be great if @tkol2022 could confirm the fixed report renders with colors against that file. Any malformed or missing data island now degrades to a console error instead of wiping out the report's color coding.

### ✅ Pre-approval checklist
- [x] I have read the CONTRIBUTING document.

### DIFF
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
@@ -1,11 +1,11 @@
 window.addEventListener('DOMContentLoaded', () => {
     const MAX_DNS_ENTRIES = 20;
 
-    const darkMode = getJsonData('dark-mode-flag') === "true";
-    const caps = getJsonData('cap-json');
-    const riskyApps = getJsonData('risky-apps-json');
-    const riskyThirdPartySPs = getJsonData('risky-third-party-sp-json');
-    const severityScoreWeights = getJsonData('severity-score-weights-json');
+    const darkMode = tryGetJsonData('dark-mode-flag') === "true";
+    const caps = tryGetJsonData('cap-json');
+    const riskyApps = tryGetJsonData('risky-apps-json');
+    const riskyThirdPartySPs = tryGetJsonData('risky-third-party-sp-json');
+    const severityScoreWeights = tryGetJsonData('severity-score-weights-json');
 
     buildExpandableTable(caps, "caps");
     buildExpandableTable(riskyApps, "riskyApps", severityScoreWeights);
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
@@ -1,11 +1,11 @@
 window.addEventListener('DOMContentLoaded', () => {
     const MAX_DNS_ENTRIES = 20;
 
-    const darkMode = tryGetJsonData('dark-mode-flag') === "true";
-    const caps = tryGetJsonData('cap-json');
-    const riskyApps = tryGetJsonData('risky-apps-json');
-    const riskyThirdPartySPs = tryGetJsonData('risky-third-party-sp-json');
-    const severityScoreWeights = tryGetJsonData('severity-score-weights-json');
+    const darkMode = getJsonData('dark-mode-flag') === "true";
+    const caps = getJsonData('cap-json');
+    const riskyApps = getJsonData('risky-apps-json');
+    const riskyThirdPartySPs = getJsonData('risky-third-party-sp-json');
+    const severityScoreWeights = getJsonData('severity-score-weights-json');
 
     buildExpandableTable(caps, "caps");
     buildExpandableTable(riskyApps, "riskyApps", severityScoreWeights);
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/Utils.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/Utils.js
@@ -16,6 +16,23 @@ const getJsonData = (id) => {
 }
 
 /**
+ * Like getJsonData, but returns null instead of throwing when the element is missing or
+ * contains invalid JSON. One bad data island must not abort the rest of report rendering
+ * (e.g. colorRows), which would leave the report without its pass/fail color coding.
+ *
+ * @param {string} id The ID of the <script> element.
+ * @returns {any} The parsed JSON data, or null if the data could not be parsed.
+ */
+const tryGetJsonData = (id) => {
+    try {
+        return getJsonData(id);
+    } catch (error) {
+        console.error(`Failed to load JSON data for "${id}":`, error);
+        return null;
+    }
+}
+
+/**
  * Ensures the input value is returned as an array.
  * 
  * @param {any} val The value to normalize.
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/Utils.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/Utils.js
@@ -1,33 +1,23 @@
 /**
- * Retrieves JSON data derived from the text content of a <script> element with the specified ID.
- * 
+ * Retrieves and parses JSON data from the text content of a <script> element with the specified ID.
+ *
+ * A missing element or invalid JSON is logged and returns null rather than throwing, so one bad
+ * data island cannot abort the rest of report rendering (e.g. colorRows), which would leave the
+ * report without its pass/fail color coding.
+ *
  * @param {string} id The ID of the <script> element.
- * @returns {any} The parsed JSON data.
- * @throws {Error} If the element with the specified ID is not found or if the JSON parsing fails.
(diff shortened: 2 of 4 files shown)


stark256-spec:
While working on this fix I noticed a separate, unrelated bug in the report scripts and kept it out of this PR to avoid scope creep.

The dark mode default flag never applies. The `dark-mode-flag` data island contains a raw `true`/`false` (from `$DarkMode.ToString().ToLower()` in CreateReport.psm1), so `JSON.parse` returns a boolean. But both IndividualReport.js and ParentReport.js compare it to a string:

```js
const darkMode = getJsonData('dark-mode-flag') === "true";
```

`true === "true"` is always false, so running with `-DarkMode $true` never defaults the report to dark mode; only the manual toggle (via sessionStorage) works. The fix would be comparing against the boolean (`=== true`) in both files.

If that sounds right to you, I am happy to open a separate small PR for it so the changes stay clean.

FollyBeachGurl:
@mitchelbaker-cisa is scuba diving

PowerShell/ScubaGear/Modules/CreateReport/scripts/Utils.js
```diff
@@ -15,6 +15,23 @@ const getJsonData = (id) => {
...
 
+/**
+ * Like getJsonData, but returns null instead of throwing when the element is missing or
+ * contains invalid JSON. One bad data island must not abort the rest of report rendering
+ * (e.g. colorRows), which would leave the report without its pass/fail color coding.
+ *
+ * @param {string} id The ID of the <script> element.
+ * @returns {any} The parsed JSON data, or null if the data could not be parsed.
+ */
+const tryGetJsonData = (id) => {
```
mitchelbaker-cisa: The new exception handling introduced in `tryGetJsonData()` can be merged into `getJsonData()` as a single function, then `tryGetJsonData` can be removed since its fairly duplicative.

```js
/**
 * Retrieves JSON data derived from the text content of a <script> element with the specified ID.
 * 
 * @param {string} id The ID of the <script> element.
 * @returns {any} The parsed JSON data.
 * @throws {Error} If the element with the specified ID is not found or if the JSON parsing fails.
 */
const getJsonData = (id) => {
    const el = document.getElementById(id);
    if (!el) {
        console.error(`Element with id "${id} not found`);
        return null;
    }
    try {
        return JSON.parse(el.textContent);
    } catch (error) {
        throw new Error(`Failed to parse JSON from element with id "${id}":`, error);
        return null;
    }
}
```

PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
```diff
@@ -1,11 +1,11 @@
 window.addEventListener('DOMContentLoaded', () => {
     const MAX_DNS_ENTRIES = 20;
 
-    const darkMode = getJsonData('dark-mode-flag') === "true";
-    const caps = getJsonData('cap-json');
-    const riskyApps = getJsonData('risky-apps-json');
-    const riskyThirdPartySPs = getJsonData('risky-third-party-sp-json');
-    const severityScoreWeights = getJsonData('severity-score-weights-json');
+    const darkMode = tryGetJsonData('dark-mode-flag') === "true";
```
mitchelbaker-cisa: These function calls can be changed back to `getJsonData()` after the update to Utils.js is made.

mitchelbaker-cisa (changes requested):
Hi @stark256-spec, thanks for submitting this fix. I tested your changes locally and can confirm the AAD report displays correctly now.

I left a few recommendations that I'd like added to this PR, let me know if you have any questions.

mitchelbaker-cisa:
> While working on this fix I noticed a separate, unrelated bug in the report scripts and kept it out of this PR to avoid scope creep.
> 
> The dark mode default flag never applies. The `dark-mode-flag` data island contains a raw `true`/`false` (from `$DarkMode.ToString().ToLower()` in CreateReport.psm1), so `JSON.parse` returns a boolean. But both IndividualReport.js and ParentReport.js compare it to a string:
> 
> ```js
> const darkMode = getJsonData('dark-mode-flag') === "true";
> ```
> 
> `true === "true"` is always false, so running with `-DarkMode $true` never defaults the report to dark mode; only the manual toggle (via sessionStorage) works. The fix would be comparing against the boolean (`=== true`) in both files.
> 
> If that sounds right to you, I am happy to open a separate small PR for it so the changes stay clean.

Yeah this definitely is a bug, good catch. I'm good with creating a separate PR, thanks!

stark256-spec:
Thanks for testing it and for the review. Consolidated as suggested — merged the exception handling into `getJsonData()` and removed `tryGetJsonData()`, and reverted the call sites back to `getJsonData()`. One small adjustment from your snippet: I kept `getJsonData()` **returning `null`** (rather than throwing) on a missing element or parse failure, since the call sites now rely on that to keep rendering the rest of the report when one data island is bad — throwing here would abort `colorRows` again, which was the original issue this PR addresses. Also fixed a stray unterminated string in the "not found" message. Let me know if you'd prefer it to throw instead.

mitchelbaker-cisa (approved):
Tested the getJsonData try/catch statement from dev tools, works as expected:

Image

And reran a test report one more time by running Invoke-SCuBACached against a ScubaResults.json file without the `risky_applications`, `severity_score_weights` keys, colorRows() ran correctly.

nward626 (approved):
## Legend

|              |    |
| ------------ | -- |
| Pass         | ✅  |
| Fail         | ❌  |
| Inconclusive | ⚠️ |

## TLDR;
1.  Invoke-Pester PowerShell/ScubaGear/Testing/Unit/PowerShell/CreateReport/ — 73/73 passing ✅
2.  Invoke-PSSA (repo PSScriptAnalyzer config) — 0 errors, 0 warnings ✅
3.  Verified tryGetJsonData degrades gracefully for all three failure modes (empty island, missing element, null island) and that buildExpandableTable(null, ...) returns early as before ✅
   
## Invoke-Pester PowerShell/ScubaGear/Testing/Unit/PowerShell/CreateReport/ — 73/73 passing ✅
All 73 tests passed 

Image

## Invoke-PSSA (repo PSScriptAnalyzer config) — 0 errors, 0 warnings ✅
PSSA returned two Information-level findings but no warnings or errors 

Image

## Verified tryGetJsonData degrades gracefully for all three failure modes (empty island, missing element, null island) and that buildExpandableTable(null, ...) returns early as before ✅
 All three failure modes and buildExpandableTable return expected outputs  

Image

PR #2282, opened 2026-07-16
visualize defender anti phishing policies

### LINKED ISSUE #2024: Add the capability to visualize Defender Anti-phishing policies in the Security Suite report
### Prerequisites

### 💡 Summary

Add the capability to visualize Defender Anti-phishing policies in the Security Suite report 

### Motivation and context
Policy group 2 is impersonation protection, which is configured by adding users to anti-phishing policies. Having a table that shows how their tenant's anti-phishing policies would help users understand why they are not passing the policies if they fail, similar to the CAP table at the end of the AAD report.

### Implementation notes
Is there any code we can reuse from the CAP tables? The structure of these tables might be quite similar.

Implementation might also be similar to https://github.com/cisagov/ScubaGear/issues/2021.

### Acceptance criteria
A table for visualizing anti-malware policies is added to the Security Suite report.

jmosby1:
#  #
  

## 🗣 Description ##
This PR adds Tables describing the anti-phish policies related to SecuritySuite Group 2. The tables are added to the end of the SecuritySuite report similar to the report tables in AAD 

It can be difficult to determine what is causing checks in SecuritySuite Group 2 without understanding the variables that are being checked. These tables provide additional context to help users track down what is missing from their policies. Closes  #2024

## 🧪 Testing ##
Run Invoke-Scuba with a configuration file (use sample-config-files/scuba_compliance.yaml as a template) that includes SensitiveUsers and PartnerDomains. You should see the relevant tables at the bottom of the SecuritySuite report

## ✅ Pre-merge checklist ##

- [x] PR/feature branch passes functional tests for relevant products, if applicable.
- [x] Feature branch has been rebased against changes from parent branch, as needed.

  Use `Update branch` button below or use [this](https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request) reference to rebase from the command line.
- [ ] Resolved all merge conflicts on branch.
- [ ] Squash all commits into one PR level commit using the `Squash and merge` button.

## ✅ Post-merge checklist ##

- [ ] Close issues resolved by this PR if the [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) did not activate.

### DIFF
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
@@ -41,6 +41,53 @@ const getPartnerDomainRows = (partnerDomains) => {
         .map(domain => ({ "Partner Domain": domain }));
 };
 
+const SAFETY_TIP_FIELDS = [
+    ["First contact", "EnableFirstContactSafetyTips"],
+    ["Similar users", "EnableSimilarUsersSafetyTips"],
+    ["Similar domains", "EnableSimilarDomainsSafetyTips"],
+    ["Unusual characters", "EnableUnusualCharactersSafetyTips"],
+    ["Via tag", "EnableViaTag"],
+    ["Unauthenticated sender", "EnableUnauthenticatedSender"]
+];
+
+const formatProtectedValues = (values) => {
+    const normalizedValues = normalizeToArray(values)
+        .map(value => String(value ?? "").trim())
+        .filter(value => value.length > 0);
+    return normalizedValues.length > 0 ? normalizedValues.join("\n") : "None";
+};
+
+const isEnabled = (value) => value === true || String(value).toLowerCase() === "true";
+
+/**
+ * Converts anti-phish policy settings into rows for the protection-policy table.
+ *
+ * @param {Array<Object>|null} antiPhishPolicies The exported anti-phish policies.
+ * @returns {Array<Object>} Unique policy rows.
+ */
+const getAntiPhishPolicyRows = (antiPhishPolicies) => {
+    const seenPolicies = new Set();
+
+    return normalizeToArray(antiPhishPolicies).reduce((rows, policy) => {
+        if (!policy || typeof policy !== "object") return rows;
+
+        const policyName = String(policy.Name ?? policy.Identity ?? "Unnamed policy").trim() || "Unnamed policy";
+        const policyKey = String(policy.Identity ?? policyName).trim().toLowerCase();
+        if (seenPolicies.has(policyKey)) return rows;
+        seenPolicies.add(policyKey);
+
+        rows.push({
+            "Policy": policyName,
+            "Users Protected": formatProtectedValues(policy.TargetedUsersToProtect),
+            "Partner Domains Protected": formatProtectedValues(policy.TargetedDomainsToProtect),
+            "Safety Indicators": SAFETY_TIP_FIELDS
+                .map(([label, field]) => `${label}: ${isEnabled(policy[field]) ? "Enabled" : "Disabled"}`)
+                .join("\n")
+        });
+        return rows;
+    }, []);
+};
+
 /**
  * Creates a simple report table that matches the static ConvertTo-Html shape
  * expected by applyScopeAttributes.
@@ -68,6 +115,7 @@ const createSecuritySuiteTable = (columns, rows, tableClass) => {
         columns.forEach(column => {
             const td = document.createElement("td");
             td.textContent = row[column] ?? "N/A";
+            td.style.whiteSpace = "pre-line";
             tr.appendChild(td);
         });
         tbody.appendChild(tr);
@@ -107,8 +155,9 @@ const appendSecuritySuiteTableSection = (parent, title, columns, rows, tableClas
  *
  * @param {Array|string|null} sensitiveUsers The configured SensitiveUsers values.
  * @param {Array|string|null} partnerDomains The configured PartnerDomains values.
+ * @param {Array<Object>|null} antiPhishPolicies The exported anti-phish policies.
  */
-const buildSecuritySuiteConfigTables = (sensitiveUsers, partnerDomains) => {
+const buildSecuritySuiteConfigTables = (sensitiveUsers, partnerDomains, antiPhishPolicies) => {
     if (sensitiveUsers === undefined || sensitiveUsers === null ||
         partnerDomains === undefined || partnerDomains === null) {
         return;
@@ -139,5 +188,14 @@ const buildSecuritySuiteConfigTables = (sensitiveUsers, partnerDomains) => {
         getPartnerDomainRows(partnerDomains),
         "securitysuite-partner-domains-table",
         "No partner domains defined in the config file."
+    );
+
+    appendSecuritySuiteTableSection(
+        section,
+        "Anti-Phish Protection Policies",
+        ["Policy", "Users Protected", "Partner Domains Protected", "Safety Indicators"],
+        getAntiPhishPolicyRows(antiPhishPolicies),
+        "securitysuite-anti-phish-policies-table",
+        "No anti-phish policies were exported."
     );
 };
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
@@ -1,0 +1,143 @@
+/**
+ * Parses a SecuritySuite SensitiveUsers config entry into table data.
+ *
+ * SensitiveUsers accepts either an email address or "Username;Email".
+ *
+ * @param {Array|string|null} sensitiveUsers The configured SensitiveUsers values.
+ * @returns {Array<Object>} The normalized table rows.
+ */
+const getSensitiveUserRows = (sensitiveUsers) => {
+    return normalizeToArray(sensitiveUsers)
+        .map(user => String(user ?? "").trim())
+        .filter(user => user.length > 0)
+        .map(user => {
+            const separatorIndex = user.indexOf(";");
+            if (separatorIndex === -1) {
+                return {
+                    "Username": "N/A",
+                    "Email": user
+                };
+            }
+
+            const username = user.slice(0, separatorIndex).trim();
+            const email = user.slice(separatorIndex + 1).trim();
+            return {
+                "Username": username || "N/A",
+                "Email": email || "N/A"
+            };
+        });
+};
+
+/**
+ * Parses SecuritySuite PartnerDomains config values into table data.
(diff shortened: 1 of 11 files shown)


adhilto:
Thanks for working on this!

Some high-level feedback:
- Could you add a add description above the "Sensitive Users" and "Partner Domains" tables? In particular, I want to make it clear to the users that those tables come from their ScubaGear config file, which may or may not reflect their actual configuration. I'm picturing something like the paragraph above the DNS tables of the EXO report that explains what those tables are.
- A key piece that is missing is who the anti phish policies apply to. The provider already collects this data:
    - For custom rules, look at the `anti_phish_rules` key. The relevant fields are `SentTo`, `SentToMemberOf`, `RecipientDomainIs`, `ExceptIfSentTo`, `ExceptIfSentToMemberOf`, and `ExceptIfRecipientDomainIs`. One interesting piece about how those fields work is they are AND not OR conditions. As in if a rule applies a group and a domain, the rule only applies to users that are in both the group and the domain. ScubaGear specifically looks for rules that have all tenant domains in the `RecipientDomainIs` field and have the other 5 fields I listed empty.
    - For the standard and strict preset policies, look at the `protection_policy_rules` key. If `SentTo`, `SentToMemberOf`, `RecipientDomainIs`, `ExceptIfSentTo`, `ExceptIfSentToMemberOf`, and `ExceptIfRecipientDomainIs` are all null, the policy applies to everyone (that's specifically what ScubaGear is looking for for the preset policies). Note that's just for the preset policies, for the custom policies, M365 doesn't allow you to have them all be null.
    - The default policies always applies to all users.
- As to how to show that data, I don't recommend listing all the users, groups, and domains. There could potentially be long lists, too much for us to display in the reports. Instead, I recommend mirroring how the CAP tables at the bottom of the AAD report shows this information. For example, you could add a "Applicability" column:
    - If the policy applies to everyone as described above, say "All users"
    - Otherwise, have a bulleted list with entries for each field that isn't null, like "Users included: 2 users", "Groups included: 2 groups", "Domains included: 2 domains", "Users excluded: 2 users", etc. Since ScubaGear is looking for all domains to be added, we could even say something like "Domains included: 2 out of 4 tenant domains" for the domains entry to help users see if a policy is failing because they didn't include all domains.
- Could you turn the data in the "Safety Indicators" fields into bulleted lists?

jmosby1:
image

Here's a screenshot of the updated code addressing Alden's comments. I'm not 100% sure if he still wanted the Users Protected and Partner Domains Protected columns, but can keep/remove those based on reviewer feedback

PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
```diff
@@ -0,0 +1,387 @@
...
+/**
+ * Creates a simple report table that matches the static ConvertTo-Html shape
+ * expected by applyScopeAttributes.
+ *
+ * @param {Array<string>} columns The table columns.
+ * @param {Array<Object>} rows The table rows.
+ * @param {string} tableClass The CSS class to add to the table.
+ * @returns {HTMLTableElement} The created table.
+ */
+const createAntiPhishExpandButton = (expanded, onClick) => {
```
mitchelbaker-cisa: The `createAntiPhishExpandButton` and `createAntiPhishTruncationButton` functions duplicate an existing pattern that exists in AADTableFunctions.js, [`createChevronIcon`](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/Modules/CreateReport/scripts/AADTableFunctions.js#L241) and [`createRowActionButton`](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/Modules/CreateReport/scripts/AADTableFunctions.js#L265), which build a button with a chevron/truncation icon and a click handler.

Since we're going to be adding more of these tables (anti-malware and anti-spam policy tables), I think it's worth refactoring the `createChevronIcon` and `createRowActionButton` functions out of AADTableFunctions.js and moving them into the top-level TableFunctions.js so they're more reusable for each individual report page.

Let me know what you think of this approach:
1. Move `createChevronIcon` and `createRowActionButton` functions out of AADTableFunctions.js and into TableFunctions.js so they're available as shared helpers.
2. Update `createRowActionButton` to always set the `aria-label`/`aria-expanded` attributes. This is a good accessibility addition you added and something I missed when first writing the `createRowActionButton` function.
3. Update `createRowActionButton` such that `rowIndex` is an optional parameter. This parameter is only used in the AAD tables for the `expandRow`/`collapseRow` functions and isn't used in the Security Suite row handlers.
4. Update `createRowActionButton` such that `expanded` is required, this is the true/false flag used to set the `aria-expanded` attribute. Since the existing `createRowActionButton` calls in AADTableFunctions.js don't have the `expanded` argument set, this would need to be added to each invocation.
5. Replace `createAntiPhishExpandButton`/`createAntiPhishTruncationButton` functions with direct calls to the shared `createRowActionButton` function.

Example for the new `CreateRowActionButton` function:

```javascript
const createRowActionButton = ({ title, className, rowIndex, expanded, onClick, contentBuilder }) => {
  const btn = document.createElement("button");
  btn.type = "button";
  btn.title = title;
  btn.setAttribute("aria-label", title);
  btn.setAttribute("aria-expanded", expanded.toString());

  if (rowIndex !== undefined) btn.rowNumber = rowIndex;
  if (className) btn.classList.add(className);
  btn.addEventListener("click", onClick);

  const content = contentBuilder();
  if (content) btn.appendChild(content);
  return btn;
};
```

PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
```diff
@@ -0,0 +1,387 @@
...
+    row.appendChild(actionCell);
+
+    columns.forEach(column => {
+        const cell = document.createElement("td");
+        appendAntiPhishPolicyCell(cell, data[column], expanded, expand);
+        row.appendChild(cell);
+    });
+};
+
+const createSecuritySuiteTable = (columns, rows, tableClass) => {
```
mitchelbaker-cisa, line 316: This is similar to AADTableFunction's [`buildExpandableTable`](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/Modules/CreateReport/scripts/AADTableFunctions.js#L284) function but making it reusable across AAD and Security Suite is scope creep. I'm thinking we can refactor this in a future issue once the anti-malware and anti-spam tables are added as well.

PowerShell/ScubaGear/Modules/CreateReport/scripts/SecuritySuiteTableFunctions.js
```diff
@@ -0,0 +1,387 @@
...
+        rows.push({
+            "Policy": policyName,
+            "Enabled": isEnabled(policy.Enabled),
+            "Applicability": getPolicyApplicability(
+                policy,
+                antiPhishRules,
+                protectionPolicyRules,
+                acceptedDomains
+            ),
+            "Users Protected": formatProtectedValues(policy.TargetedUsersToProtect),
```
mitchelbaker-cisa: Can the "Users Protected" column also output the list of users in a bulleted list, similar to what we're doing for the "Safety Indicators" column?

mitchelbaker-cisa (changes requested):
Ran ScubaGear locally and took a look at the anti-phish table generated in the Security Suite report, looks good overall. Left a few comments for your review below.

atuomit:
I find the “Users Protected” column a bit misleading. At first glance, it could be confusing to see that a policy applies to “all users” but protects “none” users. If we want to keep this field, I wonder if “Impersonation Protection” might be a clearer name.

I also saw that @mitchelbaker-cisa requested that the users be listed as bullets. What if we used bullets when there are users listed, but when the field is “None,” we simply display “None” without a bullet?

@adhilto, should we also consider adding a Priority column? I think this could be particularly helpful for configurations with multiple custom policies.

jmosby1:
Latest branch should address all comments

mitchelbaker-cisa:
Thanks for resolving my comments @jmosby1, tested both AAD/Security Suite and looks good. 

Also tested adding a 2nd protected domain with SCuBA cached, can we use a bulleted list for the domains as well? 

Screenshot 2026-08-11 at 6 43 05 AM

adhilto:
@jmosby1 Thanks for addressing my earlier comments!

A few other things:
- I second @atuomit suggestion to add a priority column
- Could you make the "Applicability" column a bulleted list too if there are multiple entries?

mitchelbaker-cisa (approved):
Tested against both Security Suite/AAD to verify refactor of the functions removed from AADTableFunctions.js -> TableFunctions.js works as expected.

I'll create a follow-up issue to investigate refactoring the table generation logic and will set it to blocked until all related issues for adding the remaining Security Suite tables are complete (anti-spam, anti-malware).

jmosby1:
@atuomit @adhilto Applicability is now bulleted except for "All Users", which remains the normal text. I also added a policy column. Interestingly the values somewhat differ from what is shown on the admin screen. Below is what you see in the tenant, but Standard Preset policy actually has a priority of 1 and Strict Preset policy has a priority of 0. These two policies only appeared on Cisaent, but just found it interesting. 
image

atuomit:
> Below is what you see in the tenant, but Standard Preset policy actually has a priority of 1 and Strict Preset policy has a priority of 0. These two policies only appeared on Cisaent, but just found it interesting.

For features that use both preset and custom security policies, policies are evaluated in the following order: Strict preset security policy → Standard preset security policy → Custom policies (in increasing numeric priority) → Default policy. In some features the default policy may also be referred to as the built-in policy, but that is not the case for the anti-phishing policies.

Could we artificially populate the Priority column to reflect the actual evaluation order? The JSON may technically return a priority of 0 for multiple policies, displaying several policies with the same priority could be confusing to users.

I also looked into why preset security policies aren’t appearing on the Anti-Phishing page for some tenants. Preset security policies can be applied to Exchange Online Protection (EOP), Defender for Office 365, or both. When a preset security policy is applied to EOP, it appears in the protection_policies_rules field of the JSON. When it is applied to Defender for Office 365, it appears in the atp_policies_rules field.

It appears that when a preset security policy is applied to Defender for Office 365 but not to EOP, the policy does not appear on the Anti-Phishing page and is not present in protection_policies_rules. However, it is still present in anti_phish_policies and may be reported as Enabled. Based on this, I think we should not rely solely on the Enabled value in anti_phish_policies when determining whether to display a preset security policy in the table. We should also check protection_policies_rules to confirm that the preset policy is applied to EOP before including it.

atuomit (approved):
Tested on all four tenants in interactive and non-interactive modes, works as expected and looks great! Thanks for all the hard work!

FollyBeachGurl:
@jmosby1 reviews are complete.

PR #2277, opened 2026-07-10
Fix EXO functional test helpers to catch and surface REST API errors

skirkpatrickMSFT:
## Description

This PR addresses intermittent EXO functional test failures across all tenants. It both catches and surfaces REST API errors that were previously silent, and eliminates the eventual consistency failures that caused intermittent test flakiness.

Previously, Invoke-EXORestMethod used Invoke-RestMethod with a bare try/catch that only said "Exchange Online API call 'X' failed" with no HTTP status code, no retry logic, and some errors silently swallowed. Now every failure is visible in the CI log with full diagnostic detail, even if the test ultimately passes via retry.

Changes:

1. Added Invoke-FunctionalExoRestRequest in Testing/Functional/Products/FunctionalTestUtils.ps1:
   - Routes EXO functional test REST calls through Invoke-FunctionalTestRestRequest for HTTP status code visibility
   - Catches and surfaces HTTP errors with Write-Information "[EXO] CmdletName failed with HTTP StatusCode..." so they appear in GitHub Actions logs
   - Retries transient HTTP errors (500, 502, 503, 504, 429) with exponential backoff (3 attempts)
   - Tolerates 404 on Get-* and Remove-* cmdlets (idempotent operations)
   - Detects API-level errors embedded in 200 OK responses (@odata.error / error fields) that were previously invisible
   - Every retry attempt and every final failure emits a Write-Information message for full CI log traceability
   - Non-transient errors (403, 400, etc.) throw immediately with the full error detail

2. Rewired Invoke-FunctionalExoCommand to use Invoke-FunctionalExoRestRequest instead of calling Invoke-EXORestMethod directly

3. Converted all 6 remaining RunScuba EXO tests to ScubaCached in Testing/Functional/Products/TestPlans/exo.testplan.yaml:
   - MS.EXO.1.1v2 (remote domain auto-forwarding)
   - MS.EXO.5.1v1 (SMTP client authentication)
   - MS.EXO.6.1v1 (contact sharing policies)
   - MS.EXO.6.2v1 (calendar sharing policies)
   - MS.EXO.7.1v1 (external sender warning rules)
   - MS.EXO.13.1v1 (mailbox auditing)

   This eliminates eventual consistency failures caused by EXO's eventually-consistent replicas, where writes to one replica are not immediately visible to reads from another.

4. Added convergence waits and replica propagation sleeps to all EXO mutation wrapper functions that previously lacked them (for future use if any tests are added back as RunScuba):
[...]

PR #2331, opened 2026-08-09
Surface REST API errors in the SecuritySuite provider instead of hiding them

### LINKED ISSUE #2319: Export-SecuritySuiteProvider hides the error message when some of the REST APIs fail
### Prerequisites

### ScubaGear Version

Latest version main branch.

### Operating System

Windows 11

### PowerShell Version

5.1

### M365 Environment and License(s)

Not environment specific.

### 🐛 Summary

The logic in Export-SecuritySuiteProvider eats the error message when any of the REST API calls mentioned below fail. This makes it impossible to troubleshoot why a problem occurred but it also causes unexpected effects such as displaying a message in the report that the tenant is missing the Defender for Office license which it actually has the license.

Affected API calls:

- Get-AtpPolicyForO365
- Get-ATPProtectionPolicyRule
- Get-DlpCompliancePolicy
- Get-DlpComplianceRule
- Get-ProtectionAlert

### Steps to reproduce

1. To simulate an error change the line of code which the cmdlet:

$ATPPolicyResult = @(Invoke-SecuritySuiteTrackedCommand -CmdletName "Get-AtpPolicyForO365" -SuppressWarning $true)

change it to:

$ATPPolicyResult = @(Invoke-SecuritySuiteTrackedCommand -CmdletName "BOGUSAPI" -SuppressWarning $true)

2. Observe that you won't see any error message in the command line output and you won't see any error in the log.
3. Look at the report and you will see the following message for some of the policies: 
**NOTE: Either you do not have sufficient permissions or your tenant does not have a license for Microsoft Defender for Office 365 Plan 1 or Plan 2, which is required for this feature.**

#### Example
This section provides screenshots from an actual execution which spawned this issue. I executed ScubaGear with securitysuite two times.

**First execution:**
When I ran ScubaGear with Invoke-Scuba -ProductNames securitysuite against an E5 tenant, for some reason I noticed the warning message in the report which indicates that the tenant was missing the Defender for Office license, but I logged into the tenant and verified that it has the license therefore the message in the report is inaccurate. 

Image

Notice that no errors show in the command output. Also notice there is no visual record that the Get-AtpPolicyForO365 API actually ran.

Image

The only clue that helped me determine that an API failed was in the JSON:

Image

**Second execution:**
This section provides a screenshot from the second execution. I executed securitysuite again just a minute after the first execution and this time the cmdlet did NOT fail so the report output the correct information.

Image

### Expected behavior

The provider should output error messages in the console output and log file. I believe the current logic does not do that because it assumes that if the REST API fails, it simply indicates that the specific API is not available in that tenant (indicating that the required license is missing). If that is the case, the code needs to be enhanced to catch the specific error which indicates there is no license and act accordingly. All other errors unrelated to the license missing should be treated as API failures and output to the console and log file.

### Output from Initialize-SCuBA (optional)

_No response_

mateenali66:
## 🗣 Description ##

Export-SecuritySuiteProvider treated any failure of the ATP and DLP cmdlets as a missing license and discarded the error, so a transient API failure produced a false missing-license note in the report with nothing in the console or log.

The provider now records each failed cmdlet's error message, treats a failure as missing-license evidence only when the message contains a 400 status or says the cmdlet is not recognized, and surfaces every other failure with the existing Write-Warning idiom plus a Write-ScubaLog entry, following the same pattern CommandTracker.TryCommand uses. Real failures stay in `securitysuite_unsuccessful_commands`, which the report already renders as an API error. Genuinely unlicensed tenants keep the prior behavior.

The 400/not-recognized classification is the load-bearing assumption. Either signal alone counts as license evidence, and I could not confirm the exact unlicensed-tenant error text against a live tenant, so review from someone with tenant access would help.

A tenant with a valid Defender license was shown as missing it whenever an ATP call failed for an unrelated reason, and the underlying error was discarded, leaving nothing to troubleshoot with. Same error-surfacing direction as #2277 took for the EXO functional test helpers. Targeted resolution #2319.

## 🧪 Testing ##

- Six new Pester unit tests in Export-SecuritySuiteProvider.Tests.ps1: license vs non-license failures for both the ATP and DLP paths, warning assertions, and JSON validity on failure.
- The full provider unit suite passes 18/18 locally after the Aug 30 rebase, which brought in new upstream tests (pwsh 7.6.4, Pester 5.9, macOS). CI's Windows PowerShell 5.1 was not run locally. The code is kept 5.1-compatible.
- PSScriptAnalyzer on both changed files: no new findings (one pre-existing Information finding on main).

- [ ] All relevant type-of-change labels added.
- [ ] All relevant project fields are set.

## ✅ Pre-merge checklist ##

- [ ] PR passed smoke test check.
- [ ] Feature branch has been rebased against changes from parent branch, as needed.
- [ ] Resolved all merge conflicts on branch.
- [ ] Squash all commits into one PR level commit using the `Squash and merge` button.

## ✅ Post-merge checklist ##

- [ ] Close issues resolved by this PR if the [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) did not activate.


sylvesterkaczmarek (commented):
Test-MissingLicenseError classifies every HTTP 400 as a missing-license signal. A malformed request or unrelated Bad Request will therefore still be hidden as “license missing” and the failed command treated as successful. Could this match a product-specific unlicensed error/code instead of status 400 generically, with a non-license 400 regression?

mateenali66:
2319 is closed as a duplicate of 2376, so the Resolves keyword on this PR points at a dead issue.
https://github.com/cisagov/ScubaGear/issues/2319

Two open issues cover this file now.
https://github.com/cisagov/ScubaGear/issues/2376
https://github.com/cisagov/ScubaGear/issues/2377

This branch already does what 2377 asks. It checks the error before it decides the license state. It also fixes part of 2376, because the Get-UnifiedAuditLogRetentionPolicy failure now reaches the console and the log instead of vanishing.

The rest of 2376 is not done here. That issue wants the two custom helpers replaced by TryCommand, and SuppressWarning removed rather than narrowed.

@tkol2022 @mitchelbaker-cisa both issues ask for an approach review before anyone starts, and 2377 says to work them together. I can take both. Do you want this PR repointed at 2377, or closed so I restart from the combined approach?

On the 400 question above, a bare 400 is enough on its own today, and that path does call AddSuccessfulCommand. It is not every 400 though. The license branch only fires when every failed cmdlet in the group looks like a license failure. A mixed 400 and 500 falls through to the API error path instead. Which codes really mean unlicensed is the open question in 2377, and I have no tenant to check the real error text against.

mateenali66:
you're right, it was matching any 400. `Test-MissingLicenseError` no longer looks at the status code, it matches an authorization refusal in the Admin API error body, so a parameter binding failure or any other Bad Request now falls through, stays in the unsuccessful command list and raises a warning instead of flipping `defender_license`. two regression tests cover that, one for the ATP cmdlets and one for DLP, and both fail on the previous commit.

one thing came with it. `Invoke-EXORestMethod` was throwing only `$_.Exception.Message`, which for `Invoke-WebRequest` is the status line and nothing else, so there was no product specific text to match on in the first place. it now carries the response body too, same approach as `Invoke-ScubaRestMethod` in https://github.com/cisagov/ScubaGear/commit/a10d56463ba12f676d8722502b1c3e203665d819.

worth flagging that I could not find a published capture of the real unlicensed response for these cmdlets, so the match is on the authorization exception the API raises when the cmdlet is outside the caller's role scope, not on a documented licence code. if you or @tkol2022 have a capture from an unlicensed tenant I will pin it to that exact code.

the sturdier option, if you'd rather, is to stop inferring the licence from the error at all and read it from `service_plans`, which the provider already pulls via `Get-MgBetaSubscribedSku`, the way `ExportAADProvider` does with `LicenseHelper`.

sylvesterkaczmarek (commented):
The missing-license classification is now based on authorization/product-specific error text rather than every HTTP 400, and the REST helper preserves the response body needed to make that distinction. The added regressions cover unrelated Bad Request failures remaining visible. My previous concern is resolved.

FollyBeachGurl:
LFT

PR #2346, opened 2026-08-13
Fix dark mode default never applying in HTML reports

stark256-spec:
Fixes the report **DarkMode** default never taking effect.

## Root cause

`CreateReport.psm1` emits the default dark mode state as an **unquoted** boolean:

```powershell
"<script type='application/json' id='dark-mode-flag'> $($DarkMode.ToString().ToLower()) </script>"
```

so the script element contains ` true ` / ` false ` (not `"true"`). `getJsonData()` runs `JSON.parse(...)` on that, returning a JavaScript **boolean**. But `IndividualReport.js` and `ParentReport.js` compared it to the **string** `"true"`:

```js
const darkMode = getJsonData('dark-mode-flag') === "true";
```

`true === "true"` is always `false`, so generating a report with DarkMode enabled produced a report whose initial state was still light mode.

```
JSON.parse(" true ")            -> true      (boolean)
true === "true"  // old  -> false   (dark mode never applies)
true === true    // new  -> true
```

## Fix

Compare against the boolean `true` in both report scripts:

- `IndividualReport.js`
- `ParentReport.js`

Two-character change, one line each; no other behavior affected. (Separate from #2258 — that PR touches `Utils.js`/`CreateReport.psm1`, not these two files.)

## Testing

Verified the parse/compare behavior directly (`JSON.parse(" true ")` is the boolean `true`, so `=== "true"` is always false and `=== true` is correct). Happy to attach a before/after screenshot of a report generated with `-DarkMode` if useful.

### DIFF
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/IndividualReport.js
@@ -1,7 +1,7 @@
 window.addEventListener('DOMContentLoaded', () => {
     const MAX_DNS_ENTRIES = 20;
 
-    const darkMode = getJsonData('dark-mode-flag') === "true";
+    const darkMode = getJsonData('dark-mode-flag') === true;
     const caps = getJsonData('cap-json');
     const riskyApps = getJsonData('risky-apps-json');
     const riskyThirdPartySPs = getJsonData('risky-third-party-sp-json');
--- a/PowerShell/ScubaGear/Modules/CreateReport/scripts/ParentReport.js
+++ b/PowerShell/ScubaGear/Modules/CreateReport/scripts/ParentReport.js
@@ -1,5 +1,5 @@
 window.addEventListener('DOMContentLoaded', () => {
-    const darkMode = getJsonData('dark-mode-flag') === "true";
+    const darkMode = getJsonData('dark-mode-flag') === true;
 
     applyScopeAttributes();
     mountDarkMode(darkMode, "Parent Report");


stark256-spec:
@mitchelbaker-cisa this is the standalone PR for the dark-mode default flag bug that came up during the #2258 review, which you suggested opening separately.

Quick recap of the root cause: `CreateReport.psm1` emits the `dark-mode-flag` element as an unquoted boolean (`$($DarkMode.ToString().ToLower())`), so `getJsonData()` `JSON.parse`s it into a JS **boolean**. `IndividualReport.js` and `ParentReport.js` compared that to the **string** `"true"`, and `true === "true"` is always false — so generating a report with `-DarkMode` still opened in light mode. Fix is a one-token change in each file (`=== "true"` → `=== true`). It's independent of #2258 (that PR touches `Utils.js`/`CreateReport.psm1`, not these two).

Tagging you since you'd already looked at it — happy to add a before/after screenshot if useful.

FollyBeachGurl:
LFT

PowerShell/ScubaGear/Modules/Orchestrator.psm1
```diff
@@ -1771,10 +1771,12 @@ function Invoke-ReportCreation {
             $ScriptsPath = Join-Path -Path $ReporterPath -ChildPath "scripts" -ErrorAction "Stop"
             $ParentReportJS = Get-Content (Join-Path -Path $ScriptsPath -ChildPath "ParentReport.js") -Raw
             $UtilsJS = Get-Content (Join-Path -Path $ScriptsPath -ChildPath "Utils.js") -Raw
+            $TableFunctionsJS = Get-Content (Join-Path -Path $ScriptsPath -ChildPath "TableFunctions.js") -Raw
```
mitchelbaker-cisa, line 2200: I ended up finding out that this TableFunctions.js script wasn't being injected into the parent HTML report correctly. 

ParentReport.js has the following function calls:

```js
const darkMode = getJsonData('dark-mode-flag') === true;
applyScopeAttributes(); // With TableFunctions.js missing this threw an error, resulting in mountDarkMode never running
mountDarkMode(darkMode, "Parent Report");
```

Readded it and committed this into your feature branch.

mitchelbaker-cisa (approved):
I replicated the issue locally against main and then confirmed your feature branch fixed the issue. See my comment below on one nuance I found while testing (resolved), but otherwise this is ready to merge.

PR #2342, opened 2026-08-13
Migrate teams to direct API calls and Enhance Functional Test Orchestrator to run faster and streamline authentication logic

tkol2022:
## 🗣 Description ##

A list of all the changes in this PR are described below:

closes #2222
closes #2335
closes #2096
closes #2314
closes #2350 
closes #2162 
closes #2201
closes #2367
closes #2345 
closes #2368

- Converted Teams Export Provider into using Teams REST APIs. 
- Added logic to mint Teams Admin API authentication tokens and Teams Unified settings API tokens.
- Converted Teams functional tests into using Teams REST APIs. Added helper functions to Teams REST APIs FunctionalTestUtils.ps1 to replace MicrosoftTeams module cmdlets.
- Enhanced Teams Rego unit tests to cover untested scenarios and ensure the logic matches the actual Rego rulesets.
- Removed all references to MicrosoftTeams PowerShell dependency module.
- Enhanced functional test orchestrator Products.Tests.ps1 to remove unnecessary calls to ScubaGear's Connect-Tenant and streamlined minting of API tokens needed for the functional tests. The only products that require tokens in the functional tests are: SharePoint, PowerPlatform, Teams and PowerBI. The other products do not modify the tenant (they use Invoke-ScubaCached) so they don't need auth tokens.
- Fixed Initialize-Msal so that it works with functional test orchestrator and mint tokens without having to make unnecessary calls to Connect-Tenant first.
- Fixed Initialize-Msal so that it actually loads the requires C# types in PowerShell 5 and 7. Previously it did not load the required types because Connect-MgGraph (which ScubaGear calls first) was actually doing that. The same approach does not work for 5 and 7 so I had to use a different technique for each.
- Refactored redundant Connect-MgGraph authentication logic into a common function Connect-GraphIfNecessary.
- Refactored Get-TenantDetail to remove redundant product specific Get-[Product]TenantDetail functions and instead use a common Get-AADTenantDetail function.
- Removed unused code files from repo.
- Enhanced helper function Invoke-ScubaRestMethod to capture the details of REST API calls failures since those were getting hidden.
- In TeamsConfig.rego changed PoliciesBlockingDefaultApps to NonCompliantPoliciesBlockingDefaultApps so that it is obvious what the set is used for.
- Changed Commandlet field values in TeamsConfig.rego to reflect the updated REST APIs that the Teams rulesets rely on.
- Enhanced Disconnect-ScubaTenant to consolidate logic and execute quickly now that authentication has changed for most products.
[...]

PR #2351, opened 2026-08-17
fix(logging): restore provider trace output after repeated runs

### LINKED ISSUE #2350: ScubaGear trace output for export providers only works during the first execution and Trace-ScubaFunction errors if running a product before AAD
### Prerequisites

### ScubaGear Version

Latest version main branch

### Operating System

Windows 11

### PowerShell Version

5.1

### M365 Environment and License(s)

Not environment specific. Powershell problem.

### 🐛 Summary
Since opening this issue, a second problem has been noted which has a very similar root cause and therefore I am documenting here so it can be rectified at the same time as the fix for problem 1.

#### Problem 1 (original manifestation of the bug)
When you open a fresh PowerShell window and execute Scubagear, it will show you the trace output for each API call inside the export providers (e.g. PowerPlatform, AAD, etc.0.
When you run ScubaGear again for the 2nd, 3rd, 4th ... N time, it no longer shows you the provider API call trace outputs.

#### Problem 2
If you run ScubaGear against any non-AAD product first and then run ScubaGear with AAD, you will get an error that the function Trace-ScubaFunction is not recognized.

### Steps to reproduce problem 1

Open a fresh PowerShell window.
Execute ScubaGear against any provider.
Observe the trace output for provider API calls.

Image

Execute ScubaGear again and observe that the API calls are no longer shown in the trace output.

Image

### Steps to reproduce problem 2
Open a fresh PowerShell window
Execute ScubaGear against powerplatform
Everything works fine, no errors

Image

Execute ScubaGear again but against aad
Notice the error related to Trace-ScubaFunction

Image

### Expected behavior

Regarding problem 1, the trace output for provider API calls should be displayed on every execution. Trace output is created by the Trace-ScubaFunction inside the ScubaLogging module.
Regarding problem 2, the provider should be able to resolve Trace-ScubaFunction from the ScubaLogging module.

### Output from Initialize-SCuBA (optional)

_No response_

breezeFur:
## Summary

Fixes #2350.

- Unload `CommandTracker` during `Remove-Resources` cleanup.
- Prevent provider modules from retaining stale `ScubaLogging` function references after a second `Invoke-SCuBA` call in the same PowerShell session.
- Add a regression test that verifies the helper module is removed during cleanup.

## Validation

- `git diff --check`
- Focused PowerShell verification passed: `CommandTracker` is loaded, `Remove-Resources` runs, and the helper is no longer loaded.
- The repository's Pester tests could not run locally because the machine has Pester 3.4.0 while this test suite uses Pester 5 operators; the existing test file fails at its pre-existing `Should -Not` assertions under Pester 3.4.0.

### DIFF
--- a/PowerShell/ScubaGear/Modules/Orchestrator.psm1
+++ b/PowerShell/ScubaGear/Modules/Orchestrator.psm1
@@ -2065,6 +2065,10 @@ function Remove-Resources {
         Remove-Module $Provider -ErrorAction "SilentlyContinue"
     }
 
+    # CommandTracker imports ScubaLogging functions into its module scope. It must
+    # be unloaded before the logging module is reloaded or providers keep stale
+    # function references on subsequent Invoke-SCuBA calls in the same session.
+    Remove-Module "CommandTracker" -ErrorAction "SilentlyContinue"
     Remove-Module "ScubaConfig" -ErrorAction "SilentlyContinue"
     Remove-Module "RunRego" -ErrorAction "SilentlyContinue"
     Remove-Module "CreateReport" -ErrorAction "SilentlyContinue"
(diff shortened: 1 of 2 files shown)


mitchelbaker-cisa:
Closing as duplicate of #2342 which is addressing the underlying issue.

Issue #2374, opened 2026-08-31
SecuritySuite 5.2 policy check does not consider priority and can produce a false Pass


tkol2022:
### Prerequisites

- [x] This issue has an informative and human-readable title.

### ScubaGear Version

Current version main branch

### Operating System

Windows 11

### PowerShell Version

5.1

### M365 Environment and License(s)

Not environment specific.

### 🐛 Summary

SecuritySuite policy 5.2 needs to consider the priority field value according to the Microsoft documentation. It currently passes the policy check even if a higher priority audit retention policy that is non-compliant is defined in the tenant. This is a false Pass scenario.

The policy check may also need to consider:

- The users who are targeted (we should probably ensure that all users are targeted with a compliant retention policy)
- The record type configured to be captured

### Steps to reproduce

I have configured a non-compliant scenario in the G5 test tenant. Just go to the Purview audit log retention policies and observe that I created a 7 day retention policy with priority 1 which is the highest priority. However the ScubaGear policy check picks up the retention policy with a priority of 2 which is configured for 1 year.

Microsoft: "A lower value indicates a higher priority. Valid priorities are numerical values between 1 and 10000. A value of 1 is the highest priority"

<img width="1436" height="318" alt="Image" src="https://github.com/user-attachments/assets/1f8bb34e-9387-47fe-8a60-b4b590c82587" />

<img width="1040" height="235" alt="Image" src="https://github.com/user-attachments/assets/2442501f-e6f4-4167-a72d-7d5ed9c94755" />

<img width="1461" height="608" alt="Image" src="https://github.com/user-attachments/assets/d4d932d0-2f07-4be6-b86a-aab437e18248" />

### Expected behavior

The priority should be considered so ScubaGear does not produce a false Pass.

tkol2022:
### Additional enhancements
I also recommend the following additional enhancements to the baseline document for 5.2 as part of this issue.

- The baseline policy title is confusing. "Audit logs SHALL be retained and searchable for a minimum of 3 months and retrievable for a minimum of 12 months." It sounds like there are three requirements. The logs shall be retained (1), searchable (2) and retrievable (3). To make it simple I would just state that the requirement is that he logs shall be retained because that is what the policy implementation instructions are about.
- Stating a value of 3 months and 12 months in the same requirement is confusing. I would simplify and state that the logs shall be retained for a minimum of 12 months.
[...]

PR #2386, opened 2026-09-04
Fix MS.SECURITYSUITE.5.2v1 false Pass when a higher-priority retention policy is non-compliant

simonyang08:
### Summary

MS.SECURITYSUITE.5.2v1 previously reported Pass whenever any retention policy in the tenant kept logs for 12 months or longer, regardless of the policy Priority. Per the Microsoft Purview documentation, a lower numeric Priority means higher precedence, so a Priority=1 SevenDays policy should outrank a Priority=2 TwelveMonths policy. Tenants that configured a high-priority short retention policy alongside a lower-priority compliant policy were incorrectly reported as compliant (false Pass).

### Changes

- `PowerShell/ScubaGear/Rego/SecuritySuiteConfig.rego`: when at least one enabled retention policy reports a numeric Priority, the highest-priority (lowest Priority value) policies now drive the result — all of them must be duration-compliant. Tenants without Priority fields keep the existing duration-only behavior. When several enabled policies tie on the smallest Priority value, all of them must comply (fail closed): Purview does not define tie semantics, and a complete rule selecting a single tied policy would abort the whole evaluation with `eval_conflict_error`.
- `Testing/Unit/Rego/SecuritySuite/SecuritySuiteBaseConfig.rego`: add `HighPriorityNonCompliantRetentionPolicy` (7 days, Priority=1) and `LowerPriorityCompliantRetentionPolicy` (12 months, Priority=2) fixtures.
- `Testing/Unit/Rego/SecuritySuite/SecuritySuiteConfig_05_test.rego`: add `test_AuditLogRetention_Incorrect_HigherPriorityNonCompliant` (fails on main, passes after fix), a symmetric compliant case, and two Priority-tie cases (both-compliant passes, one-non-compliant fails).

### Verification

- `opa test` for the SecuritySuite suite: 198/198 pass on this branch (the new incorrect-priority test fails on `main`).
- `opa check --strict` passes.

Proposed solution for #2374

### DIFF
--- a/PowerShell/ScubaGear/Rego/SecuritySuiteConfig.rego
+++ b/PowerShell/ScubaGear/Rego/SecuritySuiteConfig.rego
@@ -842,6 +842,16 @@ AdvancedAuditingLicenses contains ServicePlan.ServicePlanId if {
     ServicePlan.ServicePlanName == "M365_ADVANCED_AUDITING"
 }
 
+# Enabled retention policies that supply a numeric Priority value. Per the
+# Microsoft Purview documentation, a lower numeric Priority value indicates a
+# higher precedence, so the smallest Priority value supersedes all others.
+# See issue #2374.
+EnabledPrioritizedRetentionPolicies contains Policy if {
+    some Policy in input.unified_audit_log_retention_policies
+    not Policy.Enabled == false
+    is_number(Policy.Priority)
+}
+
 # Save audit log retention policies that retain logs for at least 12 months
 # and are not disabled.
 CompliantRetentionPolicies contains {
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
There is a partial-data false-pass case in the new priority branch. It activates as soon as at least one enabled policy has a numeric Priority, but from that point the evaluation only considers EnabledPrioritizedRetentionPolicies. Any other enabled retention policy whose Priority is missing, null or non-numeric disappears from the decision entirely.

For example, a compliant Priority=1 policy alongside an enabled SevenDays policy with no usable Priority can pass, even though the precedence of the second policy is unknown. Given that this change is specifically hardening the rule against false Pass results, I think mixed or incomplete priority data should fail closed, or otherwise be handled explicitly, rather than silently excluding those policies. A regression with one numeric-priority policy and one enabled policy lacking Priority would cover this boundary.

simonyang08:
Thank you for the careful review — you are right, and this is exactly the false-pass shape this change was meant to prevent. Fixed in fcb25d42.

**Change:** enabled retention policies without a usable numeric Priority are now collected in `EnabledUnprioritizedRetentionPolicies`. The prioritized branch of `AuditRetentionRequirementMet` only fires when that set is empty, so mixed or incomplete priority data fails closed instead of silently excluding those policies. This covers missing, null, and non-numeric Priority values.

One implementation note: `Priority` is read via `object.get(Policy, "Priority", null)` because a bare `not is_number(Policy.Priority)` does not hold for policies whose Priority key is absent under the current OPA engine, which the first cut of the fix tripped over.

**Regression coverage** (both fail closed, as requested):
- compliant `Priority=1` policy + enabled SevenDays policy lacking Priority (`test_AuditLogRetention_Incorrect_MixedPriorityDataFailsClosed`)
- same outcome when the unprioritized policy is itself duration-compliant (`test_AuditLogRetention_Incorrect_MixedPriorityAllCompliantStillFailsClosed`)

Both regressions were verified RED against the previous head (`expected false, got true` reproducing the reported false pass) and the full Rego suite passes 1062/1062 with the fix.

FollyBeachGurl:
We are closing this pull request because the associated issue is not in our current milestone and is out of scope for this release. The issue must be triaged and prioritized before implementation can begin. Please re-submit once it has been added to an active milestone. Thank you.

sylvesterkaczmarek:
Thanks for addressing the mixed-priority case in fcb25d42. The fail-closed handling resolves the concern I raised. Understood on the milestone scope and closure.

Issue #2337, opened 2026-08-11
invoke-scubacached fails in different drive working directory


DickTracyII:
### Prerequisites

- [x] This issue has an informative and human-readable title.

### ScubaGear Version

1.8.0

### Operating System

Windows 11, Server 2022

### PowerShell Version

5.1

### M365 Environment and License(s)

ALL

### 🐛 Summary

When running from anywhere C drive: It PASSES, 
However when running in same command from another drive letter (eg E:); It FAILS

### Steps to reproduce

Open PowerShell window, run:

`Invoke-ScubaCached -Productnames * -outpath e:\scuba\8_11_2026\  -OpaPath e:\scuba\opa\ -ExportProvider $false`

However when running in same command from another drive letter (eg E:); It FAILED

```powershell
CD E:\
Invoke-ScubaCached -Productnames * -outpath e:\scuba\8_11_2026\  -OpaPath e:\scuba\opa\ -ExportProvider $false
```

```
Join-Path : Cannot find path 'E:\scuba\8_11_2026\TestResults.json' because it does not exist.
At C:\Users\ritracyi\OneDrive -
Microsoft\Documents\WindowsPowerShell\Modules\ScubaGear\1.8.0\Modules\CreateReport\CreateReport.psm1:204 char:17
+ ... $FileName = Join-Path -Path $OutPath -ChildPath "$($OutRegoFileName). ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (E:\scuba\8_11_2026\TestResults.json:String) [Join-Path], ItemNotFoundEx
   ception
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.JoinPathCommand
```
`

### Expected behavior

it should allow the command to be ran from any location as long as the module exists

PR #2399, opened 2026-09-10
Resolve OPA paths to native filesystem locations

### LINKED ISSUE #2398: OPA execution fails with relative paths or FileSystem PSDrives
## Summary

On current main (`926dafdf124edf8e0c6f7fe7e4d939bb4f9d8943`), `Invoke-Rego` passes PowerShell paths directly to a native process in two cases:

- A relative `OPAPath` passes `Test-Path`, but `ProcessStartInfo.FileName` resolves it against the process working directory instead of the PowerShell location.
- `Resolve-Path.Path` can contain a custom FileSystem PSDrive prefix such as `ScubaCache:/input.json`. OPA cannot resolve that prefix.

The fallback executable path has the same PSDrive problem.

## Reproduction

Put an OPA executable, input JSON, a Rego policy and its `Utils` directory in a temporary folder. Create a FileSystem PSDrive for that folder, change to that drive, then call `Invoke-Rego` with `-OPAPath '.'` and relative input/policy paths. OPA fails to start even though the PowerShell existence checks pass. An explicit PSDrive-qualified executable path fails as well.

Reproduced with PowerShell on macOS. Related to #2337, but the original Windows physical-drive scenario has not been verified here.

## Expected behavior

Resolve the executable and all input paths to native filesystem paths before invoking OPA, preserving literal square brackets and spaces.

sylvesterkaczmarek:
## Summary

Resolve the OPA executable, input JSON, Rego policy and Utils directory to native filesystem paths. This preserves literal paths while handling relative OPA locations and FileSystem PSDrives.

Fixes #2398. Related to #2337; its original Windows physical-drive scenario still needs confirmation.

## Testing

- RunRego Pester suite: **16 passed**, including three new cases that fail before the fix.
- Real OPA 1.19.1 execution: relative executable, PSDrive-qualified executable and current-directory fallback all pass.
- PSScriptAnalyzer: no warnings or errors in either changed file.
- Tested on macOS PowerShell. Windows PowerShell 5.1 was not available locally.

### DIFF
--- a/PowerShell/ScubaGear/Modules/RunRego/RunRego.psm1
+++ b/PowerShell/ScubaGear/Modules/RunRego/RunRego.psm1
@@ -48,9 +48,11 @@ function Invoke-Rego {
         throw "Rego file not found at: $RegoFile"
     }
 
-    # LiteralPath so input/output paths containing wildcard characters (e.g. []) resolve correctly.
-    $ResolvedInputFile = (Resolve-Path -LiteralPath $InputFile -ErrorAction Stop).Path
-    $ResolvedRegoFile = (Resolve-Path -LiteralPath $RegoFile -ErrorAction Stop).Path
+    # OPA is a native process: resolve relative paths and PSDrives to filesystem paths.
+    # LiteralPath also preserves wildcard characters (e.g. []) in folder names.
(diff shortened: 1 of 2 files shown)


mitchelbaker-cisa:
Closing this PR as the associated issue #2398 is not in our current milestone backlog and needs to be triaged.

PR #2403, opened 2026-09-10
Centralize Get-HttpResponseDetails into Utility.psm1

### LINKED ISSUE #2402: Centralize Get-HttpResponseDetails into a shared module
### Prerequisites

- [ ] Assignee should meet with issue author before starting work.

### 💡 Summary

Move the HTTP response error-formatting logic (Get-HttpResponseDetails)
that currently exists as two separate, independently-maintained copies
into a single shared function in Utility.psm1.

### Motivation and context

PR #2342 fixed Get-HttpResponseDetails to correctly handle PowerShell
5.1 vs PowerShell 7 response object shapes (System.Net.HttpWebResponse
vs System.Net.Http.HttpResponseMessage), plus a case where PS 7 disposes
the response content stream before the error handler can read it.

That fix had to be applied twice, because the same logic exists as two
separate, independently-maintained copies:
- Inline in Invoke-ScubaRestMethod's catch block
  (PowerShell/ScubaGear/Modules/Utility/Utility.psm1)
- As a standalone function in the functional test tooling
  (Testing/Functional/Products/FunctionalTestUtils.ps1)

Without centralizing this, any future runtime-compatibility fix (or any
other bug in this logic) has to be found and fixed twice, in two
different files, with no guarantee both copies stay in sync. This
already happened once with the PS 7 fix in PR #2342.

### Implementation notes

1. Extract the response-formatting logic into a standalone, exported
   `Get-HttpResponseDetails` function in
   `PowerShell/ScubaGear/Modules/Utility/Utility.psm1`.
2. Update `Invoke-ScubaRestMethod` in the same file to call this
   function instead of duplicating the logic inline.
3. Utility.psm1 is already the natural home for this - every REST-based
   provider helper (EXO, PowerPlatform, SecuritySuite, SharePoint,
   Teams) already imports `Invoke-ScubaRestMethod` from there, so
   centralizing the error-formatting logic in the same module means it
   automatically covers all current and future REST-based providers.
4. Update `Testing/Functional/Products/FunctionalTestUtils.ps1` to
   import `Get-HttpResponseDetails` from `Utility.psm1` instead of
   defining its own copy. No changes needed in `Products.Tests.ps1`
   since `FunctionalTestUtils.ps1` already imports `Utility.psm1`
   directly.
5. Related work: PR #2342

### Acceptance criteria

- [ ] `Get-HttpResponseDetails` exists in exactly one place
      (`Utility.psm1`), exported for use by other modules.
- [ ] `Invoke-ScubaRestMethod` uses the shared function instead of an
      inline copy.
- [ ] `FunctionalTestUtils.ps1` uses the shared function instead of its
      own copy.
- [ ] No regressions in existing unit tests (Utility,
      EXO/PowerPlatform/SecuritySuite/SharePoint/Teams provider suites).
- [ ] Verified against real HTTP error responses on both PowerShell 5.1
      and PowerShell 7.

skirkpatrickMSFT:
## 🗣 Description

Extracts the HTTP response error-formatting logic (previously named
Get-HttpResponseDetails) out of Invoke-ScubaRestMethod's inline catch
block in PowerShell/ScubaGear/Modules/Utility/Utility.psm1 and makes it
its own exported function in the same file. Invoke-ScubaRestMethod now
calls this function instead of duplicating the logic inline.

Testing/Functional/Products/FunctionalTestUtils.ps1 no longer defines
its own copy of Get-HttpResponseDetails - it now imports the one from
Utility.psm1 instead. No changes were needed in Products.Tests.ps1,
since FunctionalTestUtils.ps1 already imports Utility.psm1 directly at
the top of the file.

## 💭 Motivation and context

Closes #2402

PR #2342 fixed Get-HttpResponseDetails to correctly handle PowerShell
5.1 vs PowerShell 7 response object shapes (System.Net.HttpWebResponse
vs System.Net.Http.HttpResponseMessage), plus a case where PS 7 disposes
the response content stream before the error handler can read it. That
fix had to be applied twice, because the same logic existed as two
separate, independently-maintained copies - once inline in
Invoke-ScubaRestMethod, and once as a standalone function in
FunctionalTestUtils.ps1.

Utility.psm1 is already the natural home for this: every REST-based
provider helper (EXO, PowerPlatform, SecuritySuite, SharePoint, Teams)
already imports Invoke-ScubaRestMethod from there, so centralizing the
error-formatting logic in the same module means any future fix only
needs to happen once and automatically covers all current and future
REST-based providers, plus the functional test tooling.

## 🧪 Testing

Verified against real HTTP requests (a 401 response with a real JSON
body from the GitHub API) on both Windows PowerShell 5.1 and PowerShell
7.4 - identical, correct output to before the extraction on both
runtimes, with the original exception type and .Response property still
preserved for the caller.

Added a new unit test file, Get-HttpResponseDetails.Tests.ps1, covering
all three response shapes the function handles: HttpResponseMessage
(PS7), HttpWebResponse (PS 5.1, via a duck-typed fake since that type
has no public constructor), and WebResponseObject (via a real
lightweight request, since that type's assembly isn't resolvable
without a prior real Invoke-WebRequest call). All 6 tests pass on both
PowerShell 5.1 and PowerShell 7.4.

Ran the full relevant existing test surface after the change: 105/105
passing across Get-Utf8NoBom.Tests.ps1, Set-Utf8NoBom.Tests.ps1,
Invoke-FunctionalTestRestRequest.Tests.ps1 (confirms Pester's Mock still
works correctly against the now-imported function instead of a locally
defined one), and the EXO/PowerPlatform/SecuritySuite/SharePoint/Teams
provider test suites.

## ✅ Pre-approval checklist

- [x] These code changes follow the ScubaGear content style guide.
- [x] Related issues these changes resolve are linked preferably via closing keywords.
- [ ] All relevant type-of-change labels added.
- [ ] All relevant project fields are set.
- [ ] All relevant repo and/or project documentation updated to reflect these changes.
- [ ] Functional tests added/updated to cover PowerShell and Rego changes.
- [ ] All automated checks (e.g., linting, static analysis, unit/smoke tests) passed.

### DIFF
--- a/PowerShell/ScubaGear/Modules/Utility/Utility.psm1
+++ b/PowerShell/ScubaGear/Modules/Utility/Utility.psm1
@@ -786,69 +786,113 @@ function Invoke-ScubaRestMethod {
         $ErrorRecord = $_
         $WebResponse = $ErrorRecord.Exception.Response
 
-        $ErrorBuffer = [System.Text.StringBuilder]::new()
-        [void]$ErrorBuffer.AppendLine("Exception Type: $($ErrorRecord.Exception.GetType().FullName)")
-        [void]$ErrorBuffer.AppendLine("Message       : $($ErrorRecord.Exception.Message)")
-
         if ($null -eq $WebResponse) {
-            [void]$ErrorBuffer.AppendLine("No HTTP response object was returned.")
-            Write-Information $ErrorBuffer.ToString() -InformationAction Continue
+            Write-Information "No HTTP response object was returned.`nException Type: $($ErrorRecord.Exception.GetType().FullName)`nMessage       : $($ErrorRecord.Exception.Message)" -InformationAction Continue
             throw
         }
 
-        # PS 5.1 throws WebException -> HttpWebResponse. PS 7+ throws HttpResponseException -> HttpResponseMessage.
-        # Compare the type name as a string rather than "-is [System.Net.Http.HttpResponseMessage]" because that
-        # assembly isn't loaded by default on PS 5.1 Desktop, and the type-literal itself fails to resolve there.
-        if ($WebResponse.GetType().FullName -eq 'System.Net.Http.HttpResponseMessage') {
-            [void]$ErrorBuffer.AppendLine("Status Code   : $([int]$WebResponse.StatusCode)")
-            [void]$ErrorBuffer.AppendLine("Status Text   : $($WebResponse.ReasonPhrase)")
-            [void]$ErrorBuffer.AppendLine("")
-            [void]$ErrorBuffer.AppendLine("=== Response Headers ===")
-            foreach ($Header in $WebResponse.Headers) {
-                [void]$ErrorBuffer.AppendLine("$($Header.Key): $($Header.Value -join ', ')")
-            }
-            [void]$ErrorBuffer.AppendLine("")
-            [void]$ErrorBuffer.AppendLine("=== Raw Response Body ===")
-            # On PS 7+, Invoke-RestMethod already reads and disposes the response content internally,
-            # so re-reading $WebResponse.Content throws "Cannot access a disposed object." The body text
-            # (when PowerShell can parse it) is instead surfaced via ErrorRecord.ErrorDetails.Message.
-            $ResponseBody = $ErrorRecord.ErrorDetails.Message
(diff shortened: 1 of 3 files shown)


FollyBeachGurl:
During standup the team decided to move this to the next sprint.

